PR #749, opened 2014-06-30
Windows setup instructions and fixes

MrBjarmastic:
Improved windows setup instructions in the wiki. Note that they are not complete and will not get mailpile to a fully working state. But they are hopefully a step in the right direction.


BjarniRunar:
I think this got taken care of during packaging. Thanks for taking the time, this was a useful reference during that work. Not sure why we forgot to say thanks and close, but doing so now!

Issue #870, opened 2014-09-03
Use the Python appdirs module to choose app data directory


BjarniRunar:
It's too late to add this dependency for the beta, but for 1.0 we should consider switching to using appdirs (https://pypi.python.org/pypi/appdirs/) instead of hand-calculating where our data lives.

PR #883, opened 2014-09-10
use of appdirs package, should fix issue #870

jschaul:
Hi! (First of all, thanks for working on Mailpile, it looks really good, I'm looking forward to it becoming a little more stable so I can start recommending it to my non-technical friends!)

I saw issue #870 and thought it was a nice small task to start with contributing, as I don't have extensive experience with larger python projects. If you can suggest other smaller python-based issues I'm also happy to work on that.

Question: the AppDirs optionally has an `appauthor` field, which may be used on windows machines for the user data directory - what the official author name of Mailpile? "Mailpile-Team"?

### DIFF
--- a/mailpile/config.py
+++ b/mailpile/config.py
@@ -1093,7 +1093,7 @@ class ConfigManager(ConfigDict):
 
         # Use platform-specific defaults
         # via https://github.com/ActiveState/appdirs
-        dirs = AppDirs("Mailpile")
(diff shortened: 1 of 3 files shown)


aviau:
It looks like it might need a rebase, but other than that it looks good to me. What do you think @BjarniRunar ?

BjarniRunar:
Just for the record: I haven't merged this, because it *moves* the Mailpile data. That will break things for our current users, some sort of migration will be necessary.

PR #1035, opened 2014-10-07
Feature: Allow running Mailpile in subdirectory 

### LINKED ISSUE #593: Allow mailpile www server to run from subdirectory e.g. /mailpile/
Proxying to a separate port is relatively simple with nginx but for instance I would like to run mailpile on the same server as roundcube in the following combination :-

https://mail.mydomain.com/ (roundcube)
https://mail.mydomain.com/mailpile (mailpile obviously)

The redirect in nginx works but then mailpile directs the browser to /in/inbox/ which then returns a 404.

It would be really nice to be able to set a url prefix so testing like this can work without ssh tunnels.

DoubleMalt:
Resolves #593 mostly. Do not merge yet, pull request is just to facilitate testing.


BjarniRunar:
Hey! How is this going? Should we be looking at what is needed to merge?  I hate having code rot...

kermit666:
Hi @BjarniRunar. Yes, @DoubleMalt and I will look into this to see how this patch works with the current master. It would really benefit us if this was merged, so that we don't have to maintain our own fork. It will probably take us some days to properly test it.

DoubleMalt:
Apparently some of my subdirectory stuff was already pulled in, at least on the frontend.
Was backend stuff also merged?

DoubleMalt:
The subdirectory stuff was just automerged. We updated it but it still has to be tested.

DoubleMalt:
Tested the current iteration of the branch and in it works fine. Please review and merge. I'm also happy to change the name of the system property (currently sys.subdirectory) if you like some other name better.

BjarniRunar:
Great stuff!

I do think we want a rename - "subdirectory" is far too ambiguous. Can we rename it sys.http_path, so it matches the http_host and http_port settings? It's also a little bit shorter. ;-)

DoubleMalt:
http_path is a great name for the property. I will adjust the pull request accordingly.

DoubleMalt:
Done! Merge at your convenience @BjarniRunar ;)

BjarniRunar:
Oops, one more thing!

The edits in */manifest.json are wrong, those are not Jinja templates, those are plugin manifests. I also am not sure about the html edits under contrib/, I think it would be best to just leave all the plugins alone.

If you can just revert those changes, I can then follow up with the fixes necessary to the plugin system so it supports this.

DoubleMalt:
Removed it. Although I'm pretty sure in sms/compose.html they are actually needed. Please check that.
Everything else was in the manifest files.

BjarniRunar:
Merged! Thanks so much for doing this. I'll add to my TODO list to go over some of the edge cases I think might have been missed.

bnvk:
Yay! Glad to see this long standing (and needed) issue merged in :grin:

DoubleMalt:
:dancers:

PR #1116, opened 2014-11-23
Don't ignore 'http' and 'https' while scanning

vrutkovs:
Remove http and https from common words: fixes empty results on links search. Note, that this would be applied to new messages only


BjarniRunar:
Sure, why not. :) Thanks for doing that!

PR #1161, opened 2014-12-22
Switching to tox and test-requirements.txt

jmvrbanac:
Allows for more consistent testing (CI and local) through the use of tox
controlled virtualenv's to execute your tests.

- Switched the Travis CI config to use the project's tox setup.

New way of executing your tests:
```
# If you need to install tox
pip install tox

# Execute tox
tox
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 mp-virtualenv
 .idea
 .*deps
+.tox
 mailpile.egg-info/
 mailpile-tmp.py
 mailpile/www/default/config.codekit
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,17 @@
 language: python
-python:
-  - "2.7"
+python: 2.7
+env:
+  - TOX_ENV=py27
+
 before_install:
   - sudo apt-get update
-  - sudo apt-get install gnupg
-install:
-  - pip install -r requirements.txt --use-mirrors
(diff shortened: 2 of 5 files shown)


BjarniRunar:
This might be a good idea, but since I have no idea what tox is, and this would break the way we currently do our development, I'm going to need some more explaining before I merge this one. :)

jmvrbanac:
Tox is pretty much the defacto standard way for Python projects to setup their testing environments. It gives a unified way that both developers and CI tools run tests against the project. 

Essentially Tox builds python virtualenvs with the required dependencies and runs jobs for things such as unit tests, coverage, and linting. 

Project/Doc Page:
[http://tox.readthedocs.org/en/latest/](http://tox.readthedocs.org/en/latest/)

There is quite a large number of Python projects that use Tox to help streamline their developer and CI workflows. I can point you to a number of those if that might be more informative.

If you have specific questions about it, I'm more than happy to help answer them, Also, I'm usually available on Freenode (jvrbanac).  :)

BjarniRunar:
I would like to merge this. Our Travis setup has been broken for ages, and if this fixes it, then hooray! If not, we're no worse off than before.

However, you're killing requirements-dev, which is used by other things. Can you just copy it?

jmvrbanac:
Re-added requirements-dev.txt as requested.

BjarniRunar:
Thanks!

PR #1171, opened 2014-12-30
Fix travis nosetest run

pall-valmundsson:
99aee250f0d467e50fce8bfd3002fcf2f4b46e20 broke travis, this should fix that.


BjarniRunar:
Thanks, sorry I didn't notice this sooner. I'm going to merge, it's the right thing even if incomplete.

pall-valmundsson:
Just so you are aware, the travis job looks like it's ok but the job actually outputs `Ran 0 tests in 0.001s`.

Issue #1105, opened 2014-11-19
Add support for nested IMAP folders


deviantintegral:
One of my mail sources runs Courier IMAP, which for folder support needs an "IMAP Prefix" specified in the client. As-is, I can't view any of the folders within Mailpile at all.

BjarniRunar:
I think this should work now, as of #1180, and e8430318afb3efccca7be0574a9a250a3dfd042a and a few more. Please re-open if I'm wrong! :-)

PR #1180, opened 2015-01-02
Walk IMAP path

kolomanschaft:
This is a fix for #1105!

_Fixes:_
- ImapMailSource walks the IMAP path and discovers all mailboxes in the tree. Before, only root level mailboxes were listed.
- The '[Gmail]' mailbox is no longer ignored since all children are now listed as well.

_Remaining issues:_
- The auto-assignment of priority tags to mailboxes might need a work over. E.g. if the Mailbox prefix is 'INBOX', the Inbox tag is assigned to all children as well.

### DIFF
--- a/mailpile/mail_source/__init__.py
+++ b/mailpile/mail_source/__init__.py
@@ -452,7 +452,7 @@ class BaseMailSource(threading.Thread):
         path = path.replace('/.', '/')
         parts = ('/' in path) and path.split('/') or path.split('\\')
         parts = [p for p in parts if not re.match(self.BORING_FOLDER_RE, p)]
-        tagname = parts.pop(-1).split('.')[0]
+        tagname = parts.pop(-1).split('.')[-1]
 #       if self.my_config.name:
 #           tagname = '%s/%s' % (self.my_config.name, tagname)
         return CleanText(tagname.replace('_', ' '),
--- a/mailpile/mail_source/imap.py
+++ b/mailpile/mail_source/imap.py
@@ -630,26 +630,31 @@ class ImapMailSource(BaseMailSource):
     def discover_mailboxes(self, unused_paths=None):
         config = self.session.config
         existing = self._existing_mailboxes()
-        discovered = []
         with self.open() as raw_conn:
-            try:
-                ok, data = self.timed_imap(raw_conn.list, '', '%')
-                while ok and len(data) >= 3:
-                    (flags, sep, path), data[:3] = data[:3], []
-                    if '[Gmail]' in path:
-                        # FIXME: Temp hack to ignore the [Gmail] thing
-                        continue
-                    path = self._fmt_path(path)
-                    if path not in existing:
-                        discovered.append((path, flags))
-            except self.CONN_ERRORS:
-                pass
-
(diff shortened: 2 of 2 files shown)


ctuffli:
Applied these patches and now see all folders under INBOX. Provider is AtMail who reportedly uses Dovecot for IMAP.

BjarniRunar:
Hrm, thanks for writing this. I wish you'd pinged me on IRC first though. :-)

This is what the 1st version of the code did. I quickly learned the error of my ways and limited the recursion for a good reason!

Basically, some IMAP servers are totally nuts and will return a user's entire home directory (with thousands of folders and files). As far as I can tell, this is a throwback to the early days of e-mail and IMAP, where an IMAP server was basically granting remote access to a Unix user's home directory, and that user may have stored his e-mail *anywhere*, and no way was an Elite Professor going to change his ways and have his home directory structure dictated to him by some lowly IMAP server...

Not only does this take a long time, it turns the Mailpile config into a multi-megabyte blob of doom, hurting performance of everything thereafter. And although this kind of thing is a weird edge case, it illustrates a DoS attack a malicious IMAP server can perform on Mailpile.

So although I agree with you that walking the tree like this is the right thing in most cases, there needs to be a limit on how many mailboxes it adds, and/or an alternate strategy for when servers throw garbage at us.

kolomanschaft:
IRC, right, sounds like a good idea. This is actually my first contribution to anything on GitHub. I'm just learning how this works in the process :-).

Regarding the IMAP path. How about simply limiting the recursion depth (say 4) and/or the total number of mailboxes (say 100). Exceeding these limitation could result in a message for the user. Maybe one could even make these parameters configurable.

BjarniRunar:
We hang out on #mailpile on freenode. You're welcome to join us! :-)

I am still not sure what the best overall strategy for this is - it is complicated by the fact that part of the mail source design goals was to periodically check for new folders/mailboxes and let the user know proactively that something new has showed up (or allow for hands-off autoconfiguration). You can't really do that if you only see a small fraction of the IMAP tree at any given time - there is no guarantee that you'll get the same 100 each time you walk the tree.

And what if the first 100 matches are all useless, and the user's mailboxes are entries 10050-10075?

One thing that needs to be explored is whether the IMAP servers give clearer indication (via flags) of what is a mailbox and what is not, so some things can be safely skipped.

And I probably need to rethink how IMAP mail source tree data is stored, so it doesn't bloat the config.

BjarniRunar:
After pondering a bit, I decided to merge this and add some sanity checks/limits on my end. Making the common case work, and having weird undefined broken behavior for people with old weird and broken IMAP servers is probably The Right Thing.

Thanks for taking the time, I look forward to more pulls from you!

Beast141:
Hi @BjarniRunar 
I just imported my first email account into mailpile and ticked tags as well on account creation, it has brought all my folders but unfortunately i see the nested ones being separated from their main folders, all dumped under inbox. Why cant i see them being nested?  However my mail is all still downloading.  My service provider is using dovecot. 

Any thoughts? If this is fixed, why arent mine showing in a nested form?

I await your reply. thanks

PR #1198, opened 2015-01-18
Update MANIFEST.in

plippie:
Fixed locale/locales typo
Added "recursive-include mailpile/contrib *"
Fixes contrib and locale directories not being added to package through setup.py


BjarniRunar:
Thanks! Awesome stuff. :)

plsng:
As this is sort of related I think I'll just post this question here: has the translation project moved somewhere else or has there just been nothing to update for a couple of months? At least it seems the source on Transifex was last updated in July of last year.

Issue #573, opened 2014-04-08
Create clever copy (jokes) while mail is indexing


bnvk:
Depending on how much email a user imports into Mailpile + how fast there machine, disk, etc are- the process can take a while. Thus, we should display clever copy and jokes along with the progress bar. I'd love to have people contribute ideas to this
- "Good things come to those who wait"
- "Make free software and be happy"
- "Most of Mailpile is made in Iceland, where much of the country believes in elves"
- "The founders of Mailpile met in a public hot tub in Reykjavik"
- "Damn kids. They're all alike."
- "We like volcanos, do you like volcanos?"
- "Email is the largest internet based social network on the planet"
- "A million hamsters are spinning their wheels right now"
- "Tapping earth for more geothermal energy"

Other ideas would be to follow various category themes like: 
- Names and quotes by political dissidents & whistle blowers
- Facts about the history of email (or mail in general)

Also useful will be phrases while generating new public keys for people
- "Fortifying encryption shields"
- "Increasing entropy & scrambling bits"
- "Digging moat, filing with alligators, fortifying walls"

PR #1214, opened 2015-01-22
Remove some jokes shown during e-mail import

jdhoek:
Some jokes that are shown during the e-mail import wizard convey a message that goes against the privacy conscious nature of Mailpile. Mailpile should probably not joke about reading the user's e-mail, because it reflects badly on the application.

For example, if you've actually ordered things from Thinkgeek during their 2011 X-mas sale, seeing **"I am pretty sure email from ThinkGeeks 2011 X-mas sale can be deleted."** does not generate a lot of trust in the software.

Definitely keep the jokes about volcanoes and the security-related lines though. :smile:


bnvk:
@jdhoek thanks for taking the time and consideration about this. I agree some messages could be alarming to some people, others will find those same messages humorous. This was actually all crowd sourced in issue #573 

I respectfully don't agree with removing all the messages you selected here. Will give this some more thought.

Another approach is designing it slightly different so it is more clearly *a joke* like making them come from little thought bubbles of the "Mailpile MailBot" or smething...

Other people please weight in on which comments you find absolutely alarming.

BjarniRunar:
Heh, first time I've actually looked at the full list of jokes. Although I agree the list probably could use some edits, for the most part it is pretty good. In particular, my take on the "reading e-mail ones" is slightly different - jokes that remind people that their e-mail contains private information, although they may be slightly jarring and uncomfortable, serve a useful purpose in that they make people think about that stuff. It's a reminder that mail clients read your e-mail, so which mail client you use, matters.

The one about "mailing wife..." has to go though, because hopefully roughly half our users will be romantically interested in men. :-)

jdhoek:
@bnvk 
> I respectfully don't agree with removing all the messages you selected here. Will give this some more thought.

No offence taken of course. That's the purpose of this pull request; I've removed the ones that seemed iffy to me, but my intent was to present this as a starting point for discussing this issue.

@BjarniRunar 
I agree that for the jokes about reading your e-mail, presentation is key. Jokes and sarcasm are notoriously hard to get right on a digital medium because of the world-wide audience with different cultures and varying degrees of English aptitude.

I like @bnvk's suggestion to separate these jokes from the more serious hints and tips:

> Another approach is designing it slightly different so it is more clearly a joke like making them come from little thought bubbles of the "Mailpile MailBot" or smething...

BjarniRunar:
@jdhoek Yes, thank you very much for bringing this to our attention and taking the time. You are quite right that this deserves consideration.

Brennan and I discussed offline (on the public #mailpile IRC channel, actually) and we both strongly feel that it's important the app remains playful. As long as jokes aren't directly offensive to anyone, I think we just need to make sure it's clear they are humour. However, as you so rightly pointed out, our users have differing backgrounds and different levels of English fluency, and I'd like to add - different levels of technical skill. We want to make it both more clear that we aren't serious, and avoid jokes that could be misunderstood to imply Mailpile is doing something it really does not.

So they either have to be so over the top / silly that they're obviously jokes, or they have to be benign when misunderstood. And of course, not obnoxious or offensive. A very few of the jokes don't quite meet those relatively objective metrics, so I'm going to rephrase/remove those now and then see where else this discussion goes...

However, I don't think a pull request is the right place for an ongoing discussion. So I'm going to close this and open a new issue for tracking "joke quality control", referencing this. I'll copy over some of the more salient points.

PR #1218, opened 2015-01-23
fix error-checking comparison bug

ohemorange:
"for _pass in ('2', '3'):" did not match with "if _pass == 3", but it does match with "if _pass == '3'".


BjarniRunar:
Thank you!

Issue #1248, opened 2015-02-21
Keyboard shortcuts j and k are backwards


MatthewWest:
j should scroll down and k should scroll up, at least if you want to be consistent with Vim, Gmail, and anyone else who does j and k for up and down.

jceb:
+1, would love to see it the other way around!

ethanwhite:
+1

MatthewWest:
I've been trying to fix this but have been looking for several hours and haven't found the keyboard shortcut map yet. Could someone point me in the right direction?

bnvk:
@MatthewWest 99% of the JS application logic lives in:

`https://github.com/mailpile/Mailpile/tree/master/mailpile/www/default/html/jsapi`

Currently the keybinding shortcuts live in `index.js` in an array. It should be straightforward what does what. Those should probably be migrated to the `global/keybindings.js` file though

MatthewWest:
Thank you! I was trying to use the minified javascript, and this is SO much easier.

PR #1262, opened 2015-03-08
Make "j" and "k" have their proper behavior

MatthewWest:
Simple fix to switch j and k behavior
Issue #1248


bnvk:
@MatthewWest thanks for the patch!

PR #1265, opened 2015-03-22
Removes "Re:" "Fw:" and so on on replies subjects

JocelynDelalande:
That PR is a work-in-progress, please don't merge it yet : it's working for reply subjects, but not yet for forwards.

To go further, I need help on how to properly add a default profile to the testing environment (in `tests/__init__.py`). @BjarniRunar ?

Also, that's my very first commit on mp's code, so any input is appreciated, especially about design choices.

Among others, I'm not totally confortable with using the jinja helper to do the job... but it does exactly the required job.

Thanks by advance.

### DIFF
--- a/mailpile/plugins/compose.py
+++ b/mailpile/plugins/compose.py
@@ -377,6 +377,17 @@ class RelativeCompose(Compose):
     _TEXT_PARTTYPES = ('text', 'quote', 'pgpsignedtext', 'pgpsecuretext',
                        'pgpverifiedtext')
 
+    _FW_REGEXP = re.compile(r'^(fwd|fw):.*', re.IGNORECASE)
+    _RE_REGEXP = re.compile(r'^(rep|re):.*', re.IGNORECASE)
+
+    @staticmethod
+    def prefix_subject(subject, prefix, prefix_regex):
+        """Avoids stacking several consecutive Fw: Re: Re: Re:"""
+        if prefix_regex.match(subject):
+            return subject
+        else:
+            return '%s %s' % (prefix, subject)
+
 
 class Reply(RelativeCompose):
     """Create reply(-all) drafts to one or more messages"""
@@ -507,7 +518,8 @@ class Reply(RelativeCompose):
 
         return (Email.Create(idx, local_id, lmbox,
                              msg_text='\n\n'.join(msg_bodies),
-                             msg_subject=('Re: %s' % ref_subjs[-1]),
+                             msg_subject=cls.prefix_subject(
+                                 ref_subjs[-1], 'Re:', cls._RE_REGEXP),
                              msg_from=headers.get('from', None),
                              msg_to=headers.get('to', []),
                              msg_cc=headers.get('cc', []),
@@ -609,7 +621,8 @@ class Forward(RelativeCompose):
 
         email = Email.Create(idx, local_id, lmbox,
                              msg_text='\n\n'.join(msg_bodies),
-                             msg_subject=('Fwd: %s' % ref_subjs[-1]),
+                             msg_subject=cls.prefix_subject(
+                                 ref_subjs[-1], 'Fwd:', cls._FW_REGEXP),
                              msg_id=msgid,
                              msg_atts=msg_atts,
                              save=(not ephemeral),
--- a/mailpile/tests/__init__.py
+++ b/mailpile/tests/__init__.py
@@ -13,12 +13,15 @@ import mailpile.util
 from mailpile.plugins.tags import AddTag, Filter
 from mailpile.crypto.gpgi import GNUPG_HOMEDIR
 from mailpile.ui import SilentInteraction
+from mailpile.vcard import AddressInfo
 
 # Pull in all the standard plugins, plus the demos.
 from mailpile.mailboxes import *
 from mailpile.plugins import *
 
 MP = None
+MY_FROM = 'team+testing@mailpile.is'
+MY_NAME = 'Mailpile Team'
 
 
 def get_mailpile_root():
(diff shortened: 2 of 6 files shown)


BjarniRunar:
Hi! Thanks for working on this.

However, this has come up before and the idea rejected - Re: and Fw: and such are part of e-mail's culture, and are there for a reason. They've been around so long that they're effectively a part of the protocol. Omitting them from Mailpile's sent mail is not an option.

The overall approach for things like this is to stay compliant with the traditions of e-mail as much as possible. If people want to slim down the user interface by omitting redundant things like this (or, preferably, translating them into more elegant UI metaphors, such as indenting in a thread view by varying amounts based on how many Re: prefixes got stripped), that is fine - but is only done at the display layer when rendering messages.

I would be open to discussing the option of not adding another Re: if there is already one present, or placing a maximum limit on them (perhaps 4 or so?) for the benefit of recipients that have clients that don't do anything useful with the prefix. But I'm not sure even that would be a net win - I've been annoyed by mail clients in the past that did "clever" things with the Re prefix which resulted in breaking the UI of *my* client which just expected a string of Re: prefixes and knew how to handle them elegantly.

This my seem like a small thing, but it impacts the formatting and information content of probably over half the mail ever composed using Mailpile, so breaking with tradition needs to be thought about carefully and supported by material arguments.

JocelynDelalande:
> Hi! Thanks for working on this.

You're welcome, quite nice code to dig in ;)

> However, this has come up before and the idea rejected - Re: and Fw: and such are part of e-mail's culture, and are there for a reason. They've been around so long that they're effectively a part of the protocol. Omitting them from Mailpile's sent mail is not an option.

Could you give me the URL of the ticket/discussion (if any) so that I can read it to avoid pulling twice the same arguments ?

> The overall approach for things like this is to stay compliant with the traditions of e-mail as much as possible. If people want to slim down the user interface by omitting redundant things like this (or, preferably, translating them into more elegant UI metaphors, such as indenting in a thread view by varying amounts based on how many Re: prefixes got stripped), that is fine - but is only done at the display layer when rendering messages.

Yep, but the issue here is IMHO more what subject I send to my recipients, not what I see.

> I would be open to discussing the option of not adding another Re: if there is already one present, 

I agree, that's the spirit. I don't want to ban *Re:* and *Fw:* from subjects ;-), just avoid stacking them to the infinite. 

Typically, two mp users speaking to each other quickly becomes a nightmare or I have to reword subjects all the time, that's not a good experience.

>- I've been annoyed by mail clients in the past that did "clever" things with the Re prefix
On my side, all the mail client i used (incl. gmail) avoid stacking Re: / Fw. tags, so IMHO, email tradition strips tags.

>  which resulted in breaking the UI of my client which just expected a string of Re: prefixes and knew how to handle them elegantly.

I disagree that `Re:` should be something parsed to add semantics to an email : `References` plays this role. I don't see the point of parsing multiple "Re:" But I'm interested in what your client did with that in the UI (which client if not impolite ?)
 
> This my seem like a small thing, but it impacts the formatting and information content of probably over half the mail ever composed using Mailpile, so breaking with tradition needs to be thought about carefully and supported by material arguments.

Sure thing. interesting discussion btw :)

BjarniRunar:
The previous discussions were all in person, I'm afraid, so we'll just have to repeat ourselves.

You are of course right that other headers - In-Reply-To and References - provide much more accurate information. I'm pretty sure that there are (or were) some shitty mail clients out there that don't add those headers, but I'm not sure we really care about them much.

I think a reasonable middle ground might be to simply not add another Re: prefix if the subject has one already. If other clients add more we'll preserve whatever they've done for continuity. I think this strikes a balance between not "polluting" the subject line, and not doing surprising things to the contents of mail.

As an somewhat contrived example, this strategy will preserve the subject "Re: Re: Re: Re: Re: prefix spam considered harmful", while your change (if I read it correctly) would break it. :-)

JocelynDelalande:
> I think a reasonable middle ground might be to simply not add another Re: prefix if the subject has one already. If other clients add more we'll preserve whatever they've done for continuity. I think this strikes a balance between not "polluting" the subject line, and not doing surprising things to the contents of mail.

From far better than nothing :), I will edit my patch toward this goal.

@BjarniRunar would you have some time to give some help on that one : 


> To go further, I need help on how to properly add a default profile to the testing environment (in tests/__init__.py). @BjarniRunar ?

JocelynDelalande:
For information, I found my answer with @BjarniRunar hint on IRC (thanks!), added in *tests/__init__.py* that kind of line : https://github.com/mailpile/Mailpile/blob/master/scripts/mailpile-test.py#L80

JocelynDelalande:
@BjarniRunar ok, reworked with proper testing and the non-destructive approach we agreed on.

Open to comments :)

JocelynDelalande:
oh sorry for failling test, strange it didn't failed runing tests locally :-/, I'l fix it this eve.

BjarniRunar:
This looks pretty good to me. If you haven't figured out the test thing by the time I next do a pull/push cycle at my end, then I'll just merge this locally and push up. Thank you!

JocelynDelalande:
> This looks pretty good to me. If you haven't figured out the test thing by the time I next do a pull/push cycle at my end, then I'll just merge this locally and push up. Thank you!

Ok, will try to do it, but no prob if I'm too slow on that and you commit the fixed tests yourself :)

JocelynDelalande:
@BjarniRunar tests are ok now :) Thanks for your help on IRC, happy to submit my first code contribution to MP :)

BjarniRunar:
Awesome! Thank you for doing this so meticulously, great stuff.

PR #1282, opened 2015-05-09
allow capital letters in http_path

DoubleMalt:


BjarniRunar:
Thanks! We might have wanted to allow even more characters, but that is a good start.

PR #1288, opened 2015-05-29
Adds a new escapejs jinja2 filter to escape templated js strings 

### LINKED ISSUE #1083: Unexpected characters in i18n strings cause JS string issues
Since I moved from e1efa08 to de50344, I can no longer click on a tag (typical ex: « maybe-spam ») and get the menu to remove/browse the tag.

Tested on Fx 24 & Chrome 33

JocelynDelalande:
fixes #1083


BjarniRunar:
That looks great, thanks for doing this!

Issue #1315, opened 2015-07-06
Example email address in account add dialog should comply with RFC 2606


nathans:
While I understand the spirit of it, I think it is best to comply with RFC 2606 email domains when displaying example email addresses.

Pic of what I am takling about:

![manning](https://cloud.githubusercontent.com/assets/206383/8515281/8e00c336-2354-11e5-96bb-619a277e3923.png)

BjarniRunar:
See discussion in pull request: https://github.com/mailpile/Mailpile/pull/1316

BjarniRunar:
I think the consensus is we like the current statement too much to change it. :-)  Thanks for filing though!

PR #1316, opened 2015-07-06
RFC 2606 compliance, bug #1315

nathans:
This should cover my bug regarding the use of "example.com" in the product.

### DIFF
--- a/mailpile/www/default/html/contacts/add/index.html
+++ b/mailpile/www/default/html/contacts/add/index.html
@@ -9,7 +9,7 @@
       <label>{{_("Name")}}</label>
       <input type="text" name="name" data-type="name" class="contact-add-name" value="" placeholder="Chelsea Manning">
       <label>{{_("E-mail")}}</label>
-      <input type="text" name="email" data-type="email" class="contact-add-email" value="" placeholder="chelsea@manning.com">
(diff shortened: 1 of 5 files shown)


BjarniRunar:
I understand the sentiment, but if we're going to pick nits... well your replacements are too long! :-) They're not fully visible because the input field is too small.

Without resizing the input fields, we can choose between RFC compliance or Chelsea, we can't really have both. I think I'd rather make the political statement than comply with the RFC, but I'm quite willing to hear arguments either way.

DoubleMalt:
+1 for political statement and humor. I think at this stage of the project
these things should take precedence over pedantic RFC compliance.

bnvk:
@nathans thanks for the contribution + suggestion. However, I tend to agree with @BjarniRunar @DoubleMalt and like having a little bit of political fun with our naming as well.

BjarniRunar:
I think the question is: do any other similar political references fit in the box and end in @example.com ?

bnvk:
To that end, we could do `c.manning@example.com` or `chelsea.m@example.com` not sure if those fit in the fields or not, but maybe!

nathans:
I'll go ahead and close this pull request since it won't be merged at any rate. You can take or leave the suggestion, and I can even make a fresh pull request if desired. Just comment on the issue I opened if that is desired. Thanks.

BjarniRunar:
Thanks @nathans !

PR #1319, opened 2015-07-08
E-Mail migration

cricsus:
I think a tool as yandex.com and mail.ru provides would be nice. Meaning a tool to download mails from another mail server into the own (maybe just moved) one.  So if someone just moved servers and started using Mailpile, he could still access the earlier received and sent email through the new server and nothing would be lost.


BjarniRunar:
Why is this a pull request? I think you clicked the wrong button... closing. :-)

cricsus:
Lol thnk for telling. Wanted to put it as a suggestion. Haha

PR #1320, opened 2015-07-08
IMAP STARTTLS support

manuel-io:
Some error-handling is required.

### DIFF
--- a/mailpile/defaults.py
+++ b/mailpile/defaults.py
@@ -155,6 +155,7 @@ CONFIG_RULES = {
         'auth_type':       (_('Authentication scheme'), str, 'password'),
         'host':            (_('Host'), str, ''),
         'port':            (_('Port'), int, 993),
+        'force_starttls':  (_('STARTTLS'), bool, True),
         'keepalive':       (_('Keep server connections alive'), bool, False),
         'discovery':       (_('Mailbox discovery policy'), False, {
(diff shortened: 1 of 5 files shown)


BjarniRunar:
Thanks for doing this!

Not merging quite yet because I'm not sure the way you're doing the config is how I want it done; for SMTP there are three "protocols" defined for the different auth styles (clear-text, SMTP over SSL or SMTP with STARTLS) and I had planned to use the same pattern for IMAP and POP3.

Would you like to update your PR to use that style, or shall I pull this down and tweak that myself?

manuel-io:
I have seen the SMTP way. My intention was to do the same with IMAP but found out it's better this way. If there is already a IMAP connection you can parse the capabilities string like:

    if 'LOGINDISABLED' in self.capabilities:
      self.starttls()

    if 'STARTTLS' in self.capabilities:
      self.starttls()

Throw only errors if STARTTLS is enforced …

If you like to change it - do it. The IMAP/STARTTLS string is really long and you will have to adjust the CSS too. I don't know how to do this.

BjarniRunar:
I merged this locally and made the changes I wanted; avoiding the extra config variable and using the protocol similarly to how SMTP works. I kept the settings dialog unchanged, and added opportunistic encryption - if STARTTLS is available it will be used even if the user did not request it.

Tested against yandex.ru, seems to work fine.

Thanks a lot for your help with this!

PR #1328, opened 2015-07-14
add ehlo if necessary on smtp connection start & fix non mac local mail sources

DoubleMalt:
Just two little one line fixes


BjarniRunar:
Thank you!

PR #1335, opened 2015-07-21
Sidebar CSS fixes

bnvk:
Since dev is departing a bit from where I left things off / how I was implementing them, i'm going to get in the habit of pull requests for anything beyond the most trivial one line change, etc...

I made a few minor fixes here for the sidebar that had been irking me.

- fixed X overflow scrollbar
- made long tag names ellipses
- fixed hidden tag (under bottom bar)

Here's some screenshots of the ellipses on long tag names. I couldn't get the CSS to behave 100% ideally. Namely, if there is no "unread" count, it gets truncated too short and has unused space. I think the ellipses are an improvement over just chopping the tag mid word (sometimes).

screen shot 2015-07-21 at 10 39 45 am
screen shot 2015-07-21 at 10 42 19 am


karlp:
presumably mouse over shows the whole name?

bnvk:
@karlp yes, it shows the standard browser title tag on hover. We could make it show more fancy tooltips if so desired.

Doh! Don't accept this yet @BjarniRunar I just realized it messes up the expand / collapse icons. Back to the code line!

BjarniRunar:
Thanks for doing this! And thanks for making it a PR so I can review. I'm very much in favour of this change, so let me know if you get it to state you are happy with.

bnvk:
Ok. I've got the collapsible triangles fixed for all display densities. However, in even further testing, when the following three conditions are met, there is a less than desirable outcome:

- Tag name is too long and overflows
- Tag has unread messages
- Tag has subtags

screen shot 2015-07-22 at 9 59 58 am

Making the CSS handle all 3 conditions and ellipses is going to be some significant refactoring. We could however make a concession like don't display number of unread msgs, or eliminate the triangle and make the whole "Parent Tag" click to expand / collapse. This would require adding a link to the bottom of the sub-tags saying "View Contents of All" or something... if we really care about that view.

BjarniRunar:
Would it make sense to just move the triangle to the left side instead of the right, so you have [triangle][icon][name...][count] instead of [icon][name...][count][triangle] ?

Either way, this is an improvement so here goes the merge! :-)

bnvk:
I don't think [triangle][icon][name...][count] would work as it will still suffer the case of the [name...] value being insanely truncated. Additionally, the icons / triangles on the left will create a jagged / unpleasing look of:

- [icon][name]
- [icon][name][number]
- [triangle][icon][name...][number]
- [icon][name...]

I'm sure there is some better way to approach this such as make the number count (when in a collapsible item) have a background behind it, or perhaps just a hover effect for the numbers. I'll give it more thought. Thanks for accepting this!

BjarniRunar:
Truncation is unavoidable, the only solution to that is tooltips, which we have already. Users can rename tags if it bugs them. I don't consider this a problem, the space constraints are there for good reason I think.

The counts vs. triangles issue is however a problem for using the app; both are important signals to the user and the triangle is critical for accessing subtags.

When I suggest moving the triangle to the left, I envision it slightly overlapping the icon and invading the margin, but at a fixed position (position: absolute; margin-left: -NNpx). The positioning of the icons should remain unchanged, so icons and names will still line up neatly.

If the triangle is after the icon in the DOM tree, it'll automatically overwrite and capture focus when clicked, so everything *should* work relatively easily.

Making the triangle 33%-overlap the icon and 66%-invade the margin seems like a balance that could work. Since the triangles are all a similar shade to the background, it might not even be too ugly. ;-)

PR #1347, opened 2015-07-24
Prevent browsers leaking referrer headers

vtduncan:
This change is to prevent browsers sending Referrer headers from Mailpile.

It fixes the scenario where the user clicks a link in a plain-text email and their browser sends a link pointing to their Mailpile instance (and thread ID) to an external site. I see that HTML emails are already contained in a sourceless iframe and don't leak referrers—that's great!

The meta tag used here is part of the [Referrer Policy spec](https://w3c.github.io/webappsec/specs/referrer-policy/) and already works in Chrome and Firefox.

Thanks for working on Mailpile!


bnvk:
Thanks for adding this. Good call!

vtduncan:
Thanks for merging, @bnvk.

PR #1358, opened 2015-07-31
Avoid passing unicode to imaplib during login

### LINKED ISSUE #903: Error testing new sending route: "character mapping must return integer, None or unicode"
I'm stuck trying to get started with the beta release. Every time I try to add a new sending route and click "Test Route", it says "Error Connecting". On the console, this traceback appears:

```
  File "/home/ke/install/Mailpile-0.4.0/mailpile/plugins/setup_magic.py", line 927, in setup_command
    test_only=True, test_route=route))
  File "/home/ke/install/Mailpile-0.4.0/mailpile/smtp_client.py", line 312, in SendMail
    RunTimed(30, sm_startup)
  File "/home/ke/install/Mailpile-0.4.0/mailpile/util.py", line 713, in RunTimed
    raise exception[0]
TypeError: character mapping must return integer, None or unicode
```

Might it have something to do with my password? It contains special characters, but only ASCII ones.

perriv:
The imaplib library in python 2.7.9 does not automatically quote unicode types: https://hg.python.org/cpython/file/2.7/Lib/imaplib.py#l1069, so the user gets an 'Invalid username or password' error if his IMAP username contains a slash or any other character requiring the string to be quoted. This is because the username in the configuration is of type unicode.

Here's an example session showing the failure in imaplib when logging into an exchange server through DavMail:

```python
$ python
Python 2.7.9 (default, Apr  2 2015, 15:33:21)
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import imaplib
>>> from getpass import getpass
>>> m = imaplib.IMAP4('localhost', 1143)
>>> m.login('AD\\perriv', getpass())
Password:
('OK', ['Authenticated'])
>>> m = imaplib.IMAP4('localhost', 1143)
>>> m.login(u'AD\\perriv', getpass())
Password:
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/imaplib.py", line 520, in login
    raise self.error(dat[-1])
imaplib.error: LOGIN failed
```

Based off of the code that went into https://github.com/mailpile/Mailpile/commit/422aff53c919ca791bd29a4257acab5982298df2 that fixed #903.

### DIFF
--- a/mailpile/mail_source/imap.py
+++ b/mailpile/mail_source/imap.py
@@ -683,9 +683,9 @@ class ImapMailSource(BaseMailSource):
 
             try:
                 ok, data = self.timed_imap(conn.login,
-                                           my_config.username,
-                                           my_config.password)
-            except IMAP4.error:
+                                           my_config.username.encode('utf-8'),
+                                           my_config.password.encode('utf-8'))
+            except IMAP4.error, UnicodeDecodeError:
                 ok = False
             if not ok:
                 ev['error'] = ['auth', _('Invalid username or password')]


BjarniRunar:
Good catch, thanks for doing this!

I think there's a syntax error in your modification to the except line; the tuple of caught exceptions needs a paren around it.  I don't think that will catch UnicodeDecodeErrors, instead it'll set a local variable named UnicodeDecodeError when an IMAP exception is raised.

BjarniRunar:
I am going to merge this locally and clean up. Thanks again!

perriv:
Ok sounds good. No problem, happy to do it.

PR #1381, opened 2015-08-16
Changed all msgid instances of vcard to VCard

comradekingu:


BjarniRunar:
Hello! Thanks for attempting this - however you're editing the translation file directly, not the source code. That file is auto-generated from the python, HTML and Javascript. If you do this your changes will not have any effect and will ultimately get overwritten by our automated build process.

I'm sorry you have wasted your time on this, but I *did* warn you about this on IRC. I guess I didn't communicate clearly enough!

If you look in the .po and .pot files, there are lines like this one:

     #: mailpile/app.py:73

What that means, is the string was extracted from line 73, in the file `mailpile/app.py`. That is where any edits to the source string have to take place. Editing .po and .pot files will not work.

I hope this clarifies. :-)

PR #1388, opened 2015-08-24
Killed'em all (those typos)

### LINKED ISSUE #1387: Translatable strings typo collection
Typo in both XML key and string.
- Unthreaded %d messaages [messages]
- Failed to loaded metadata index [load]
- Failed to started the web server [start]
- Displayed settings [Display settings?]
- Server rejected recpient [recipient]
- Required paramter missing. Required parameters are: %s [parameter]
- Enabling spambayes autotagger [SpamBayes]
- Found key in DNSPKA [DNS PKA]
- Parts of this message has a signature that are invalid or bad ['Part' or 'have'?]
- Watch out, the signature was made with a key that has beenrevoked- this is not a good thing [been revoked]
- Filters are a very powerful aspect of Mailpile that is not yet available to non techie users. However, if you are technically inclined [there's obviously missing something at the end]
- Error accesing your encryption key [accessing]
- Dont look now, we are copying things [don't]
- Damn kids. Theyre all alike. [They're]
- All of your config settings & passwords are encrypted with AES 256 [AES-256]

RiseT:
(Almost) fixes #1387

### DIFF
--- a/mailpile/commands.py
+++ b/mailpile/commands.py
@@ -1503,7 +1503,7 @@ class RunWWW(Command):
                 return self._success(_('Started the web server on %s'
                                        ) % http_url)
         else:
-            return self._error(_('Failed to started the web server'))
+            return self._error(_('Failed to start the web server'))
 
 
 class WritePID(Command):
--- a/mailpile/commands.py
+++ b/mailpile/commands.py
@@ -2172,7 +2172,7 @@ class ConfigPrint(Command):
                                    'access_denied': access_denied
                                })
         else:
-            return self._success(_('Display settings'), result=result)
+            return self._success(_('Displayed settings'), result=result)
 
 
 class ConfigureMailboxes(Command):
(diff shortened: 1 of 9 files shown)


mailpile/commands.py
```diff
@@ -2172,7 +2172,7 @@ def command(self):
                                    'access_denied': access_denied
                                })
         else:
-            return self._success(_('Displayed settings'), result=result)
+            return self._success(_('Display settings'), result=result)
```
BjarniRunar: Hmm.. the message is telling the user what happened, thus the past tense. If that is wrong, we have a lot more messages that need changing - for consistency this should stay unchanged IMO.

BjarniRunar:
This is great, thank you! I only had one comment (see above) - do you want me to merge this and fix that myself before regenerating the messages, or do you want to update the PR?

RiseT:
@BjarniRunar I see. In this case 'Displayed settings' is correct, of course.

I've undone that change, but now it's failing again. Possibly because of the vCard commit you've just done?

BjarniRunar:
Its entirely possible the tests are testing the stability of the strings themselves in some cases. I'll pull this down and regenerate the .pot, and fix any tests needed in the process. Thanks for doing this!

PR #1389, opened 2015-08-24
Hunting for vcard, VCard & vCard

### LINKED ISSUE #1385: vcard, VCard & vCard
Mailpiles currently uses the spelling
- 'vcard' and
- 'VCard'.

Both are somewhat incorrect - 'vCard' is the official spelling.

RiseT:
Fixes #1385 

This corrects **only** the translatable strings. Documentation, class names and whatnot are left unchanged (there are just too many), although they suffer from the same lack of consistency.

### DIFF
--- a/mailpile/commands.py
+++ b/mailpile/commands.py
@@ -1179,7 +1179,7 @@ class Rescan(Command):
             self._run_rescan_command(session)
 
         if args and args[0].lower().startswith('vcards'):
-            return self._success(_('Rescanned vcards'),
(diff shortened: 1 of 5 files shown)


RiseT:
Wtf failed here... No idea. Just changing the case within a couple of strings shouldn't cause that much of chaos. Well - it least I've tried.

BjarniRunar:
I think Travis is being silly. If not, I'll fix locally - I've reviewed the changes twice and can't see anything wrong. Thank you!

PR #1396, opened 2015-09-04
Improved Dockerfile

dbrgn:
- Comments
- .dockerignore
- Added spambayes dependency
- Fig is now docker-compose
- Volume mapping is not really necessary in docker-compose.yml


DoubleMalt:
I'm not sure about the change in the docker-compose file. Mailpile does not pick up changes in the python code.
It would help for tinkering with the frontend code, though. Maybe add this as comment.

dbrgn:
Well, it depends whether the Dockerfile should primarily be used for dev or for deployment. I think the latter should be the target. For development, people can create their own volume mappings.

DoubleMalt:
If the latter is the target, the volume is unnecessary, as the ADD command already adds the files to the image on build.

dbrgn:
Exactly.

DoubleMalt:
Ok, my bad. Somehow thought you had added the volume.

Maybe add a docker-compose.dev.yml. But that does not concern this pull request, that I wholeheartedly support.

dbrgn:
Right now the docker-compose file is not really needed anyways. It only makes sense when running and linking multiple containers.

It does the port mapping for you though, so I think we can keep it.

DoubleMalt:
lgtm

DoubleMalt:
:+1:

PR #1399, opened 2015-09-09
debian-dev: add 'node' to list of packages to install

sycoso:
node is needed to run make js (and thus make all)


sycoso:
Stupid me, it has to be nodejs, not node

PR #1400, opened 2015-09-10
locale: Add missing \ to escape \n

sycoso:
There were a few \n instead of \\n in the translation (es_419 and nn_NO)


BjarniRunar:
Thanks for pointing this out!  However, this stuff needs to be fixed in Transifex (if it is still broken there), otherwise it'll just get clobbered next time we merge their changes.

BjarniRunar:
Hey @largestprime , could you takek a look at that?

Issue #1393, opened 2015-08-27
using Mozilla Autoconfiguration for determining mail server settings


ferriswheel:
just installed and now testing mailpile. 
the auto setup did not find my imap server settings.

i have used the mozilla autoconf setup for many years for myself and tech support clients.
mozilla created this method as an additional means of helping thunderbird users.
it does require a subdomain in the form of autoconfig.<domain name> and a couple of files on a webserver at that url.
no other involvement from mozilla and no secrets are kept in the config xml file. 

see  https://wiki.mozilla.org/Thunderbird:Autoconfiguration
and  https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration

regards

sycoso:
FYI: I'm trying to implement that :)

BjarniRunar:
@sycoso excellent!

The autodetection code actually has FIXMEs in place for this, in `mailpile/plugins/setup_magic.py`, which is where most of the autoconfig magic takes place. Please feel free to find me on IRC (BjarniRunar on #mailpile on Freenode) if you have any questions at all.

sycoso:
This issue should be closed, right?

ferriswheel:
hello,

i tested the master branch and found that moz autoconfig did not work.
i suspect that the port number specified in my config.xml file was ignored.
i suggest using all the the values in the xml file to give a user the control they require.
 

regards

sycoso:
What exactly do you mean by "did not work"?
I just tested it and it seems to take the port from the config…

ferriswheel:
On Wed, 30 Sep 2015 08:04:24 -0700
Jannis <notifications@github.com> wrote:

> What exactly do you mean by "did not work"?
> I just tested it and it seems to take the port from the config…
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/mailpile/Mailpile/issues/1393#issuecomment-144439287


hello,


i am using non-standard ports.
on one email server i have multiple instances of imapd,  one per domain.
each domain is assigned a different port number.
thunderbird responds appropriately by reading the port number from the config.xml file.

i am hoping mailpile, in the next few months, is working sufficiently well to start trials in a community group.
mailpile would become the recommended and supported email application with a potential 19K user base.
       

regards

sycoso:
Oh, I think I get it… do you have multiple <emailProvider>s in your config.xml? It's just that I've never seen that before.

That sounds great :)

[...]

PR #1404, opened 2015-09-16
Add support for Mozilla domain autoconfig

sycoso:
- it supports both the autoconfig subdomain and the .well-known folder
- the domain (with ssl) is queried before ISPDB,
  mx1 (and domain without ssl) is queried after ISPDB
- NOTE: Actually we want mx1 before ISPDB but DNS is easily manipulated


sycoso:
Fixes #1393

PR #1405, opened 2015-09-17
Fix problem with stale docker cache after rebuild

DoubleMalt:


DoubleMalt:
For reference:

https://github.com/docker/docker/issues/3313

BjarniRunar:
Thank you! And hello from pyCon UK :)

PR #1418, opened 2015-09-28
Fix Vagrantfile and scripts/vagrant_bootstrap.sh to actually produce working Mailpile installation

hazelesque:
As per conversation in IRC, I had a bit of trouble getting this to work in Vagrant, and I had to make some changes to the Vagrantfile and scripts/vagrant_bootstrap.sh to get "vagrant up" to spin up a working Mailpile installation.

This PR is my attempt to tidy this change up for submission, but I am new to this project, so apologies in advance if this is all wrong...

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,8 +3,7 @@
 
 Vagrant.configure("2") do |config|
 
-  config.vm.box     = 'debian7'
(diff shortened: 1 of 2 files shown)


BjarniRunar:
Thank you for doing this! It looks good to me (but be warned I don't use Vagrant).

PR #1420, opened 2015-10-02
Fix/improve arch-dev target in Makefile

sycoso:
* add --needed to pacman command to only install packages that are not already present
* add npm to pacman dependencies
* install spambayes via pacman (instead of AUR) since it's available in the repositories now
* fix URL for yuicompressor
* install base-dev before compiling yuicompressor (if not already present)

NOTE: one may also want to use the nodejs LESS compiler instead of the ruby LESS compiler since that's already in the arch repositories (and maybe even removes the ruby dependencies?). I don't know if the ruby compiler is better than the nodejs version…

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ arch-dev:
 	cd $$TMPDIR; \
 	pacman -Qs '^yuicompressor$$' > /dev/null; \
 	if [ $$? -ne 0 ]; then \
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Taking your word for it, as I do not use Arch. Thank you!

sycoso:
Thanks :)

What about the ruby less vs. nodejs less compiler? Do you know if they are interchangeable? And whether ruby is needed anywhere else?

PR #1422, opened 2015-10-10
Install bower dependencies before building less

JocelynDelalande:
That's required because of less-elements dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,7 @@ less-loop: less-compiler
         done
 
 less-compiler:
+	bower install
 	@cp scripts/less-compiler.in scripts/less-compiler.mk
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Hmm, this should be done by the 'make less' target. I don't want it done in the less-compiler recipe, as that will slow down the less-loop significantly. Is the 'make less' rule broken?

JocelynDelalande:
> Hmm, this should be done by the 'make less' target.

Are you sure ? "make less" is called by "less-loop"

BjarniRunar:
Ah, I got it backwards, you're right. Mergymergymerge. Thank you!

PR #1423, opened 2015-10-10
Rename default.css to default.min.css

JocelynDelalande:
It's a minfied/obfuscated CSS, and it's quite anoying when you want to grep it,
using a different extension allows to exclude it from your grep.


BjarniRunar:
Gonna reject this one as I've stopped obfuscating the CSS as of the latest build. :-)

Instead I've implemented gzip compression in the HTTP daemon which should accomplish the same thing quite a bit more without the ugly.

JocelynDelalande:
@BjarniRunar fair enough :)

PR #1424, opened 2015-10-10
Fix copy/paste from mail body in thread view

JocelynDelalande:
Implemented using `<br />` for displaying line returns instead of relying on
`white-space: pre-wrap` for plaintext mail rendering.

Thread rendering is exactly the same than before (pixel-level checked), except,
copy/paste will now preserve line breaks in pasted text.

Could seem relevant to implement this fix with `<p>` instead of `<br />` but it
doesn't give the exact same rendering.


BjarniRunar:
Interesting! Let's try it. ;-)

PR #1430, opened 2015-10-12
Removed tooltip animation hovering a message tag

### LINKED ISSUE #1427: Remove the animation on tag tooltip
The tooltip on tag icons (search view) is flying arround in a shiny but distracting and non-functional way, We have to kill it :)

JocelynDelalande:
In search view, qtip2 has a default to position the tooltip with an animation
from "default" to "choosen" position. That's weird.
Thanks http://stackoverflow.com/a/26672882/1377500

fix #1427


BjarniRunar:
Awesome, thank you!

Issue #1242, opened 2015-02-04
Search term with "@" causes "internal error"


JocelynDelalande:
Try a search with "`@foo`" in the web interface and you will get an HTTP 500.

BjarniRunar:
That's not a bug, that's a feature!

Well, actually it is a bug. Thanks for filing, this is the CLI interface and the web interface disagreeing, and we should make sure that no input into the search box ever causes a 500.

Andrew-Wang-IB45:
I have tried a search with "@foo" in the web interface and did not receive such a message. I tested this on Ubuntu 18.4.2 LTS using Mailpile version 1.0.0rc5.
![Mailpile Issue 1242 Tested](https://user-images.githubusercontent.com/37257276/65378993-fa0af680-dc8e-11e9-8dec-be1e47eb9506.png)

BjarniRunar:
Thanks @Andrew-Wang-IB45, you are correct. This was fixed.

PR #1436, opened 2015-11-05
Issue #1242 Search term with "@" causes "internal error"

camelaissani:


camelaissani:
I don't know if the unit tests were successful before my fix :( 
I reverted my changes replayed the tests and finally I have the same errors.
Could you confirm?

BjarniRunar:
Thanks for this, but I don't think this is the right way to handle the problem. I'll have a think about how to deal with this more gracefully.

camelaissani:
You're welcome. Let me know if I can help on this. May be with few advices I will able to finalize the fix?

PR #1439, opened 2015-11-06
Replace all occurrences of e-mail with email

simonsmiley:


simonsmiley:
It looks like the tests are wrong.
If I understood it right, the first failure is about many emails contain the phrase twitter. In that case it's 4 emails not three
the second would be about how many emails there are total. counting the mbx files there are 11

BjarniRunar:
Uhm... I prefer e-mail.

simonsmiley:
Even the IETF require email 
https://en.wikipedia.org/wiki/Email#Spelling

BjarniRunar:
That link also says that e-mail is the most common spelling anywhere there is an actual editor involved or a style guide. Seems it's geeks vs. literate types, and I'm going to side with the literate types for now - especially since this would break a bunch of translations for something that basically boils down to being a matter of taste. Thanks though. :-)

comradekingu:
When the source says email, e* makes its way into translation. Seeing as that is the real decision, and it being erroneous in just about every other language, "e-mail" is the better option.

PR #1449, opened 2015-11-06
Fix the mailpile tests

simonsmiley:
I am pretty sure that the tests were wrong...


JocelynDelalande:
@simonsmiley :-1:  tests are ok, but code is wrong, Could you check by yourself and tell me what do you think ?

See #1450 and #1451

BjarniRunar:
I broke this in my half-finished threading work. I'll double check and either merge or fix this at my end. Thank you for doing this.

JocelynDelalande:
Are you sure this fix is legit ? Why the reported count should have changed ?

PR #1452, opened 2015-11-14
Fix crash replying to a subject-less message

JocelynDelalande:
I think I introduced it in abb3a9168e492c81891961f34307b5f53737bd6f (for #1265)


JocelynDelalande:
Do *not* merge yet, I noticed that subject of empty-subject replied message is "None", which is better than a crash, but not the right thing to do.

JocelynDelalande:
Fixed, ready for merge on my side.

PR #1454, opened 2015-11-24
stop hang by introducing timeout (9s) to probe port function

largestprime:


BjarniRunar:
Regarding our chat yesterday:

    Python 2.7.6 (default, Jun 22 2015, 17:58:13) [GCC 4.8.2] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import socket
    >>> c = socket.create_connection(('192.168.0.99', 99), timeout=2)
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python2.7/socket.py", line 571, in create_connection
        raise err
    socket.timeout: timed out
    >>> 

So we know this throws socket.timeout - however, we are not explicitly catching that in the Except section. We do however catch socket.error. Is that enough? It should be if timeout inherits from error:

    >>> isinstance(socket.timeout(), socket.error)
    True

So we should be good. Testing verifies:

    >>> try:
    ...     c = socket.create_connection(('192.168.0.99', 99), timeout=2)
    ... except socket.error:
    ...     print 'woops'
    ... 
    woops

So I'm going to merge this as-is, I don't think it is causing the thread crashes we were looking for yesterday.  If you could double-check that you get the same results from those code snippets on your mac's Python, that would be great.

largestprime:
sorry this took so long, the place I stayed at on my first night
inexplicably didn't have wifi. (although, now that I think about it,
since we're going for timeouts, it would have still been ok).

yes. exactly the same stuff comes up in my mac's python.

-- 
Daniel Yeow
+45 31 32 28 84
http://www.danielyeow.com
http://darkroom.danielyeow.com

"To save the world requires faith and courage: faith in reason, and
courage to proclaim what reason shows to be true."   ~ Bertrand Russell

PR #1456, opened 2015-12-04
adding test for mailpile.config._GPGKeyCheck

infinite-Joy:
Please let me know if I can help with this.


BjarniRunar:
Thanks for doing this!

PR #1458, opened 2015-12-11
prevent double ssl wrap to ensure ISPDB info is properly retrieved within sensible timeout

largestprime:
the python 2.7.9/10 thing with the ssl/tls wrapping in the connection broker
some matching list data type thing from ages ago
and the seconds -> milliseconds timeout adjustment


BjarniRunar:
Mergymergymerge.

PR #1460, opened 2015-12-14
added tests test_sha512_512k test_sha512_512kCheck

infinite-Joy:


infinite-Joy:
Hi, Can you please clarify when the mailpile.smtp_client.sha512_512kCheck(random_challenge, bits, random_solution) would evaluate to true?

infinite-Joy:
does the build fail with the same error message in other pull requests as well. I have seen the same error before also. Any suggestions for its resolution??

BjarniRunar:
Hi! That code (the sha256_512k stuff) is a largely failed experimental and I don't think it should be further entrenched by adding test - it should probably be deleted! :-)

Thank you for doing this though! Sorry you wasted your time on this one.

PR #1464, opened 2015-12-21
move debian folder to packages/debian_docker

### LINKED ISSUE #1463: please move the Debian folder
Hello!

I am a Debian Developper and I intend to start packaging Mailpile and upload it to unstable or experimental.

Could you guys move the debian folder somewhere else? Its getting in my way :)

I'd like to refer you to the upstream guide, that says the following:

> Some projects include a rough /debian directory among source files to ease bleeding-edge package compilation and installation on debian (and derived) systems. While this is a good effort, it is better to leave it out of the final tarball as it can interfere with debian's own packaging effort. Keeping it only in your VCS repository is usually a much saner default if it lives in a specific packaging branch

Keeping it on a second branch would be the best.

aviau:
Closes #1463 

This will make my job easier. Thank you in advance.


BjarniRunar:
Thanks for looking at this! I'd like to chat with you about the packaging work you're doing, any chance you'll be on IRC at some point?

PR #1466, opened 2015-12-21
removed unused file

aviau:
This is just skeleton's example file. Its not used.

I want it removed because it contains privacy breaches (it loads javascript on the internet).


BjarniRunar:
I think this one I will reject; the file isn't used by the actual Mailpile UI, but may have value as in-place sample code & documentation for developers. It's harmless and potentially useful, so let's keep it for now.

Issue #1465, opened 2015-12-21
remove all minified javascript from this repository


aviau:
To help with packaging, remove the following files:

- ``` ./mailpile/www/default/js/libraries.min.js```
- ```./mailpile/www/default/js/plupload.full.min.js```
- ```./mailpile/www/default/js/jquery.min.js```

they can be replaced with non-minified JS. They will be minified at build time.

BjarniRunar:
Noted; not doing this just yet, as it will mean changing our documentation to warn people they need to build those files - we currently check in compiled JS and CSS as a convenience for people testing from Github.

That doesn't mean we'll never do this... just not sure now is the time.

aviau:
Solved in #1470

PR #1470, opened 2015-12-21
remove some minified js libs

aviau:
this is part of #1465

This removes some embedded minified js libs, which will be minified at build time instead.

I'll come up with a solution for libraries.js in another PR.


BjarniRunar:
I'm going to hold off on this for a bit; changing this means we also need to update our installation procedures, which assume the compiled files are checked in and folks don't need a build step after git pull.

aviau:
I am willing to update the docs.

Is there anything else than https://github.com/mailpile/Mailpile/wiki/Getting-started and its sub-pages?

BjarniRunar:
Dragging my feet a bit.. is it possible to exclude the minified JS from the tarball you build, instead of removing it from git and adding to .gitignore?

I like the other changes you've made (and would like to merge them!), but I am quite fond of the fact that folks can git pull and then just run the app, without installing a whole bunch of dev tools.

aviau:
Fair enough. I'll even allow you to include it in the tarball and I'll exclude it from the tarball when downloading it. This way we get the best of both worlds - minified js for those that want it and souce for the others.

PR #1472, opened 2015-12-22
Added tarball target

aviau:
This is my proposition for creating the source tarball.

It contains the minified ```libraries.min.js```, but it also includes the ```bower_components``` folder, so in Debian I can re-create ```libraries.min.js```.

The transifex files are missing. I don't exactly know where to pull them from and where they go.


BjarniRunar:
Very nice, thanks for doing this!

PR #1474, opened 2015-12-22
adding test_VCardLine_args_too_long to test_vcard

infinite-Joy:
seems like the new line gets inserted at a fixed position. Shouldnt it be at the start or the end of a word. Something like what happens in a normal word processor like MS Word?


BjarniRunar:
VCards aren't really human readable, so where the line gets broken doesn't matter and isn't worth complicating the code for. Thanks for the test!

PR #1483, opened 2015-12-28
prevent crash on unexisting file

aviau:
At first install, the htaccess file does not exist. The script would crash trying to remove it.


aviau:
@BjarniRunar Looks like you caught back in 5121a3669d4ebc3ffa1cf83154c5908dfe427fde

PR #1485, opened 2015-12-28
Don't run FIX_PERMS_SCRIPT when not needed

aviau:
We shouldn't have to run FIX_PERMS_SCRIPT here.

Even if we did, we are going to get permissions denied because there is no point in trying to gain ownership to files that are not already ours.


BjarniRunar:
This is needed - the script is designed to also be run from the CLI by root to do admin tasks on a multi-user install. I expect that role to grow in the future, so we want to keep this stuff around.

PR #1486, opened 2015-12-28
Reworked deb factory

aviau:
This is a rework of the ```make dpkg``` command. Just like before, ```make dpkg``` will output deb packages in ```dist/```.

I also took the opportunity to include some of my packaging work. From now on, I'll open a PR from time to time to update the ```packages/debian``` folder.

If you want to build nightlies, all you have to do is write a script that clones this repostory and does ```make dpkg```. I'll take care of the rest, making sure the packaging works.


BjarniRunar:
This looks pretty great!

Note that we'd missed a requirement in requirements.txt (will be fixing on next push) - Mailpile uses the PIL library (python-pilow) for thumbnail generation

PR #1487, opened 2015-12-28
Make Composer contact selecting hover color slighty darker

### LINKED ISSUE #1200: Make Composer contact selecting hover color slighty darker

comzeradd:
Just a small contribution during 32c3 hackathon to familiarize myself with mailpile's codebase :)

This fixes #1200

### DIFF
--- a/mailpile/www/default/css/default.css
+++ b/mailpile/www/default/css/default.css
@@ -3358,7 +3358,7 @@ Version: 3.4.2 Timestamp: Mon Aug 12 15:04:12 PDT 2013
   clear: both;
 }
 .select2-results .select2-highlighted {
-  background: #f6f6f6;
+  background: #e6e6e6;
   color: #333333;
 }
 .select2-results li em {


BjarniRunar:
Hey, thanks for this and thanks for taking part in the hackathon!

However, I can't commit this as it is - default.css is generated using less, the source files are in mailpile/www/default/less/ ... you need to track down where those lines come from and make the change there.

BjarniRunar:
Just a heads' up - things have moved around a bit. The mailpile/www/default/ folder has moved to shared-data/default-theme/

comzeradd:
It seems that this property is inherited from select2 library. I'm trying to figure out how you build the final css files, but the documentation around frontend doesn't help. The interesting thing is that select2 has different properties for this specific element than the one present at the final `default.css`.

BjarniRunar:
I think you can just put that in the app/composer.less (or something like that) - it'll override the select2 defaults. To rebuild the CSS, run 'make less' from the root of the Mailpile source folder.

PR #1494, opened 2015-12-31
Moved static out of python module

aviau:
Morning,

This pull request moves static outside of the python module. We now support installing Mailpile multiple ways.

Running ```python setup.py install```:
- as root:  will install static in ```/usr/local/share/mailpile/```
- from a virtual env: will install static inside the virtual env

Running ```python setup.py develop```:
- will symlink everything in the virtual env, allowing you to edit the files directly in the repository.

Mailpile will try to load ```<sys.prefix>/local/share/mailpile``` by default. If its not present, it will fall back to ```<sys.prefix>/share/mailpile```.

The Debian package will be able to install static in ```/usr/share``` by setting --prefix=/usr or something similar.

This is great because all install schemes including virtual envs and develop are supported, without the need for extra configuration. 

setuptool's data_files is a,  [damn](http://blog.codekills.net/2011/07/15/lies,-more-lies-and-python-packaging-documentation-on--package_data-/), [mess](https://www.youtube.com/watch?v=o-UbWsO9rZk), so I make use of pbr.


BjarniRunar:
This is pretty good, I think the only thing it breaks is the ability to just git pull and ./mp directly (assuming all requirements have been installed already).  I think I can just tweak the mp script to fix that, it's already making other similar changes to support that. Am going to re-read and will most likely then merge.

Don't we need to do the same thing for the locale/ data ?

aviau:
>  I think I can just tweak the mp script to fix that

I just did it.

> Don't we need to do the same thing for the locale/ data

I will in a future PR.

BjarniRunar:
Warning: I renamed things again... :-)

PR #1496, opened 2015-12-31
/usr/lib/cgi-bin/mailpile/admin.cgi is a symlink again

### LINKED ISSUE #1495: mailpile-admin.py: cannot find executable when in dev mode
I broke something in mailpile-admin.py:

```
'mailpile': (distutils.spawn.find_executable('mailpile') or os.path.join(mp_root, 'mp')),
```

the second part no longer works because I copy the script instead of symlinking.

```
 'cp "%(mailpile-admin)s" /usr/lib/cgi-bin/mailpile/admin.cgi',
```

Just switching back to symlinking it will break in "full-install-mode" becasue followsymlinks is disabled in ```/usr/lib/cgi-bin/``` on Debian-based distros. 

For it to work, we have to revert to symlink and make sure the owner of the symlink is the same as the owner of the script.

I'll fix this issue myself while I work on #1478

aviau:
Closes #1495

### DIFF
--- a/scripts/mailpile-admin.py
+++ b/scripts/mailpile-admin.py
@@ -46,16 +46,25 @@ MAILPILE_DELETE_SCRIPT = [
 
 
 INSTALL_APACHE_SCRIPT = [
+    # Install dependencies
     '"%(packager)s" install screen expect',
+    # Enable apache modules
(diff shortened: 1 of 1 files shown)


aviau:
closing this in favor of #1497

PR #1499, opened 2016-01-03
adding test_is8bitstring

infinite-Joy:
The name seems to be wrong. Seems to be it should be isNot8bitstring?


BjarniRunar:
Yes, that name does seem odd - not sure it's worth changing though. The generator is simply copied from Python's stdlib, we needed to modify it to avoid some misfeatures that were breaking generation of signed e-mails... thanks for doing this!

PR #1500, opened 2016-01-03
modify fonts disclamer

aviau:
The fonts are licensed under a free license, but this disclamer seemed to indicate otherwise.


BjarniRunar:
Done upstream, just need to recompile the less

aviau:
thx <3

Issue #1366, opened 2015-08-04
Need to build signed Mac OS X packages


BjarniRunar:
The Mac packages do not run on recent versions of OS X without disabling all sorts of scary sounding code signing checks which Apple have switched on by default.

Obviously we don't want to tell users to do things which degrade their security, so we're going to have to figure out how to sign our app packages properly.  This is a showstopper for 1.0 on the Mac.

NoahAndrews:
While it would certainly be ideal to sign the OS X builds, it is possible to run unsigned code without disabling Gatekeeper. If you access the Security and Privacy System Preferences pane after attempting to run an unsigned program, there will be a button allowing you to launch the program, which will whitelist the application for Gatekeeper.

AngusHardie:
You can also right click on the app in question and choose open 

Apple has a page about it:
https://support.apple.com/en-gb/HT202491

This isn't as good as getting the thing signed, but is relatively easy to explain and technical users may already know it.

NoahAndrews:
I'm pretty busy, but I'm taking a look, starting with following the Mac build instructions. Why is the brew-package script written for and tested on such an ancient version of OS X?

NoahAndrews:
I'm beginning by making that build script compatible with the current version of OS X.

NoahAndrews:
As for the actual task of signing the OS X packages, that will need to involve paying Apple a $99/year fee for their Developer Program. If I sign up on behalf of Mailpile, can I be reimbursed once I hand the credentials over to you?

NoahAndrews:
Is there a reason that the brew-package script uses symlinks 1.2 instead of 1.4? 1.2 has a makefile incompatible with the current version of OS X.

BjarniRunar:
Hi @NoahAndrews! Thanks for looking at this; it is most appreciated.

I don't think you can (or should) sign up on behalf of Mailpile, if they even allow that that would be a rather major security flaw in and of itself. What I was hoping for here, was enhancements to the script and some docs so once Mailpile has an app ID, we can make our own builds.

The reason the build is/was for an ancient version of Mac OS X, was to guarantee folks can install the app on ancient versions of Mac OS X. Developers live in a special imaginary world where everyone upgrades their laptops all the time, I had a desire to target non-technical users and make something that would actually work on older machines too, so I used an older box for both testing and packaging. At this point I've conceded that I'm too resource constrained to do everything perfectly, so if volunteers step up that don't value this I won't push the issue. :-)

[...]

PR #1503, opened 2016-01-03
Made brew-package.sh compatible with recent OS X releases

NoahAndrews:
The script now uses symlinks 1.4 instead of 1.2. I am unable to test it on any versions of OS X other than 10.11. However, it should still work.


NoahAndrews:
Is that merge commit going to be a problem?

BjarniRunar:
Thanks you for working on this!

NoahAndrews:
Glad to! Please keep an eye on issue #1366, I'm using it as a place to ask questions about the Mac version.

PR #1507, opened 2016-01-04
removed unexisting jquery link

aviau:
jquery is now in libraries.min.js


BjarniRunar:
I think more is needed - it looks to me like libraries.min.js isn't loaded :)

aviau:
Then it works without it.

aviau:
@BjarniRunar done

PR #1520, opened 2016-01-06
moved locale data

aviau:
Moved locales to shared-data/

Setting my preference to another language didn't seem to do anything, before or after my changes. So I could not reliably test this.


BjarniRunar:
I'm rejecting this, but only because I'm going to do the same thing on my local tree - I've been cleaning up and rejiggering locale stuff and don't feel like dealing with the conflicts. :-)

BjarniRunar:
See 2e9d3e1fcb41b82c9b933d2451f9f59bba6d1729 :)

PR #1524, opened 2016-01-08
Moved multipile to /usr/share/mailpile/multipile

aviau:
```/var/lib``` is for user content, and these files shouldn't change.


BjarniRunar:
Are we sure about that? Isn't this a bit like default content in /var/www/, something that the user may want to tweak & skin for their own needs?

aviau:
The contents in ```/var/www``` is really a placeholder, its not really part of the web server application.

In our case, its a working application. ```/var/lib``` is for user-uploaded content or application-generated files.

The login screen might be tweaked and skinned, but the rest of the applicaiton too.

PR #1530, opened 2016-01-11
check that we are running from a virtual env AND installed in it

### LINKED ISSUE #1526: Unable to start setup
After following the instructions for getting started on Linux (Ubuntu 15.10), creating and using a virtualenv, etc., I can't get Mailpile to work. Instead, I get this error:
```
Template not found
```
The URL is http://localhost:33411/auth/login/?_path=%2F, redirected from http://localhost:33411/

I tried running "setup" (which seemed to succeed) in the shell, but nothing changed. I'm relatively sure that I followed all the directions correctly. Is there a problem with the current master, or is something else going on? I can't find any logs created by Mailpile, so it's hard to know what's going on.

Thanks!

aviau:
closes #1526

Checking that we executing in a virtual env is not enough to assume that we are installed in a virtual env. It could be that we are only using the venv to run dependencies that and we are running from the repository itself (as we do in the Getting Started in the wiki).

### DIFF
--- a/mailpile/config.py
+++ b/mailpile/config.py
@@ -1237,9 +1237,11 @@ class ConfigManager(ConfigDict):
         if env_share is not None:
             return env_share
 
-        # Check if we have been installed in a virtual env
+        # Check if we are running in a virtual env
         # http://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv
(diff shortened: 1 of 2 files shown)


BjarniRunar:
Thanks for doing this. :-)

rmenessec:
See my new comments on #1526.

PR #1533, opened 2016-01-12
corrected test_BoolCheck_exception

infinite-Joy:
3 gives truth output for test test_BoolCheck_exception because of the logic
    try:
        # Floats are fun! :-P
        return (abs(float(txt)) >= 0.00001)
    except (ValueError, TypeError):
        pass
in line 435 inside util.py
if 3 should raise value error then this part should be corrected

### DIFF
--- a/mailpile/config.py
+++ b/mailpile/config.py
@@ -242,10 +242,9 @@ def _MakeCheck(pcls, name, comment, rules):
         _COMMENT = comment
     return Checker
 
-
 def _BoolCheck(value):
     """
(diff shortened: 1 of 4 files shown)


BjarniRunar:
This PR contains a bunch of unrelated changes (other tests tweaked), and incorrect changes (the conversion of default return to None).  I don't understand what you're trying to accomplish here!

infinite-Joy:
Yes too many things seem to be going on here and in retrospect the conversion of default return to None seems wrong. I will break them up and submit a fresh request. Closing this one.

PR #1534, opened 2016-01-12
Moved mailpile-admin inside python module [WIP]

aviau:
Work in progress, not fully tested. Just opening the PR so that you know what I'm doing and don't move around those files for now.

This PR is an attempt at solving a few things:

- fix ```get_mailpile_shared_datadir```: It didn't work because ```__file__``` no longer refers to something within the mailpile install. We can't use it to find mailpile.
- remove ```get_mailpile_shared_datadir``` code duplication
- include mailpile-admin in the PATH (as requested)

This PR allows us to run multipile from a virtual env:
```
virtualenv env
source env/bin/activate
pip install -r requirements.txt
mailpile-admin --configure-apache
```

We can also run multipile from /usr/local:
```
sudo python setup.py install
mailpile-admin --configure-apache
```

Edit: The last thing that I would like to do is find a way to let users run multipile directly from the repository.

### DIFF
--- a/shared-data/multipile/mailpile-admin.py
+++ b/shared-data/multipile/mailpile-admin.py
@@ -357,7 +357,8 @@ def get_os_settings(args):
         'webroot': args.webroot,
         'rewritemap': args.rewritemap,
         'mailpile': (args.mailpile
-                     or distutils.spawn.find_executable('mailpile')
+                     or distutils.spawn.find_executable(
+                         'mailpile', os.path.join(sys.prefix, 'bin') + ':' + os.environ.get('PATH'))
                      or os.path.join(mp_root, 'mp')),
         'mailpile-admin': os.path.realpath(sys.argv[0]),
         'mailpile-theme': (args.mailpile_theme


BjarniRunar:
OK... um, fwiw, I disagree with this. If I agreed with it, I wouldn't have duplicated the code that was marked as duplicated for a reason... :-)

BjarniRunar:
To further clarify; the Mailpile codebase, for better or worth, is pretty tightly integrated - you load one piece of it and all sorts of other stuff comes with it. CGI programs are started up over and over and over again and it's preferable that they be relatively lightweight. We can't have both lightweight and pulls-in-all-of-Mailpile, so I opted for light-weight.

aviau:
Okay, then the CGI script has to find another way to locate Mailpile because ```get_mailpile_shared_datadir``` won't work.

Maybe ```malpile-admin``` could leave a trace after setting up apache? Something in multipile.rc that will allow us to find mailpile again?

aviau:
I have updated my PR to locate the mailpile bin inside the virtual env, but I have not yet thought of a way to ensure ```mailpile-admin``` is ran from the same interpreter as ```---configure-mailpile```.

aviau:
I'll open a new PR soon.

PR #1535, opened 2016-01-13
Moved multipile to www folder + allow interpreter configuration

aviau:
This solves a few things:
- Moved multipile to ```/usr/share/mailpile/multipile/www/```. Debian policy suggests a ```www``` subfolder. (not absolutely needed, but its a good thing to do.)
- Allow python interpreter configuration: Whatever interpreter is used to run ```--configure-apache``` will be used to run the CGI script. This allows us to find the mailpile executable in a virtual env, no matter where it is on the system.

It should work everywhere already, but I'll wait for your feedback before I test all cases.


BjarniRunar:
Moving things to www/ seems fine. Two quibbles:

 - I'd not move the sample rc file into www/ as it does not belong there.
 - It seems like modifying the contents of /usr/share/mailpile/multipile/www/ (or the equivalent install target) like this violates the rule of thumb that /usr/ should be treated as read-only except by package managers. I assume your .deb package will ship an admin.cgi file itself?

Regarding the find-the-virtualenv problem, it is indeed vexing, but does this really solve it? I am not very experienced with virtualenvs and their quirks, so excuse me if I'm asking silly questions - but does finding the right interpretor/binary/path actually solve the problem that the environment is all wrong when Mailpile is launched by mailpile-admin.py? Is it possible that what mailpile-admin.py is doing in CGI mode may simply be incompatible with virtualenvs and we shouldn't be bending over backwards and complicating things to support it?

We might want to discuss this on IRC!

aviau:
> I assume your .deb package will ship an admin.cgi file itself?

Indeed, mailpile-admin shouldn't be used to install things on ```/usr``` (atleast we shouldn't recommend using it for that), it should be used to install things in a virtual env or in ```/usr/local```.

> Regarding the find-the-virtualenv problem, it is indeed vexing, but does this really solve it?

Yeah.

BjarniRunar:
OK, thanks for clarifying (here and on IRC)! Two requests:

1. stop moving the config sample into www/
2. enhance the cleanup rules in the Makefile to remove the generated admin.cgi

...and then I should be happy to merge when you're confident this is working in the common cases. :-)

aviau:
> stop moving the config sample into www/

Done.

> enhance the cleanup rules in the Makefile to remove the generated admin.cgi

Done.

> when you're confident this is working in the common cases. :-)

Tested with the following cases:

- __virtualenv:__
```
sudo su
virtualenv --system-site-packages env
source env/bin/activate
pip install -r requirements.txt
python setup.py install
(as root) env/share/mailpile/multipile/mailpile-admin.py --configure-apache
```

Watch out, you must be root when running mailpile-admin. Just sudo won't inherit your environment and won't launch the proper interpreter. In other words, sudo won't take your venv ito account.  When we write a multipile install procedure, we should document this.

- __from source:__
```
sudo scripts/mailpile-admin --configure-apache
```


- __usr/local__:
```
sudo python setup.py install
sudo /usr/local/share/mailpile/multipile/mailpile-admin.py --configure-apache
```

BjarniRunar:
Thank you! :-)

PR #1543, opened 2016-01-16
Windows-friendly CatchUnixSignals

quondammelody:
signal.SIGQUIT, signal.SIGUSR1 are not available on nt systems so it crashes on launch. This allows mailpile to launch


BjarniRunar:
Thank you!

PR #1545, opened 2016-01-16
Add blackbox functional tests in tox

JocelynDelalande:


JocelynDelalande:
(CI will fail because till #1544  gets merged, thus)

PR #1549, opened 2016-01-20
dont run make clean in setup.py

aviau:
this interferes with my work.

For example, make clean removes the locales, which I am trying to install.
It also removes the bower_components files, which I may want to keep if I am another distro than Debian and want to embed the javascript.

You may want to move things in the mrproper target instead, but  I think its easier to leave it up to whoever is installing.


aviau:
Nevermind, ```python setup.py clean`` still calls ``make clean``. so we have to fix the clean target.

PR #1550, opened 2016-01-20
Moved things out of clean target

aviau:
this interferes with my work.

Setup.py calls ```make clean```, so locales are deleted.

Also, I'm calling clean during packaging and it removes bower_components files, which are included in the tarball and I may want to keep if I decide to embed a library.

And now that I think about it, I think ```git reset --hard && git clean -dfx``` in mrproper is kinda dangerous. We should replace it with something else in a future PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,18 +97,18 @@ pytests:
 	@echo
 
 clean:
-	@rm -rf shared-data/locale/?? shared-data/locale/??[_@]*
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Hmm, okay. This means I'll need to do a bit of extra work when I finish up the i18n build stuff, but for now, okdoke.

Issue #1004, opened 2014-09-24
Fix our support for bare PGP-encrypted attachments


bnvk:
Currently if a Mailpile user is sent an attachment encrypted using PGP, in an otherwise unencrypted message (not using PGP/MIME), we do not automatically decrypt the attachment for the user. This should be fixed.

Relates to #733

JackDca:
With the above mentioned commits, bare PGP encrypted and/or signed attachments will be replaced by their unwrapped and the processed versions.

There's still work to be done - some attached files do not appear as icons at the bottom of the message.  This appears to be related to the mime type of the file and/or the success/failure of the crypto processing.

Example: It appears that attached files with mime type application/pgp-encrypted don't generate icons.  This is correct since according to RFC3156 that mime type is for the **preamble** to an encrypted mime part so the user doesn't need to see it.  (The **body** of the encrypted file is typically application/octet-stream.)

Unfortunately the mime type may be assigned by the sender's operating system based only on the file name extension and may be wrong.  Debian Squeeze/Thunderbird/Enigmail assigns application/pgp-encrypted to any file ending in ".gpg" which is doubly wrong: first, gpg uses that extension for both signed and encrypted files; second, the file is the body, not the preamble as mandated by RFC3156.

Example:  Enigmail gives an attached armored (.asc) key file the mime type application/pgp-keys if the user attaches it using the "Attach My Public Key" button on the UI, but gives it mime type text/plain if the user attaches it using the general purpose "Attach" button.

Another example:  It appears that encrypted attachments for which encryption **fails** (e.g. encrypted to some unknown key) sometimes do not generate any icon below the message.  So the user can't save it as a file and work on it outside of Mailpile, in fact he may not even know it is there.

I can see at least three ways to deal with this:
- Display an icon for **every** attachment;
- Have the ParseMessage() function analyze each attachment, using both the assigned mime type and the content, and pass on the type and/or whether it should be displayed;
- Have ParseMessage() fix incorrect mime types, at least in common or important cases.
Of course none of these solutions is without disadvantages.

If backend work is needed to implement a solution, either one of the above or one I didn't think of, I'd like to help.

BjarniRunar:
Work on this has mostly moved to #1789. You are correct that attachment handling *in general* needs some improvements, but as they aren't PGP specific I'm going to close this and file a separate issue for that. Thanks @JackDca  :-)

Issue #701, opened 2014-05-26
Encrypted public key should be decrypted and properly imported


bnvk:
Not exactly sure how to name this bug. @smari said in IRC that it is "Fails to detect / decrypt (for importing) of ASCII-armored keys" 

A key was emailed to me `0xKEYID.asc.pgp` and I was not able to import the key file into my Mac keychain downloaded by Mailpile as it was ASCII-armored. 

I had to manually decrypted via CLI with `gpg --decrypt` this undesirable behavior for downloading of keys!

Additionally, this type of attachment was not being detected as a "key to import" the mime was `application/octet-stream` we should perform some sort analysis on this attachment!

JackDca:
With the push mentioned above an encrypted PGP key attachment is properly decrypted.
It is still difficult to import because clicking the "Import Key" icon displayed below the message just displays "Eventually this will import a PGP key to a contact" but does nothing.
Is it possible for that icon to be set up to behave exactly like the "Import Key" icon that appears when a key search is done on a "To" contact?

JackDca:
P.S.  If this requires some work on the Python code, I would like to help.  JS and HTML i can't help with.

PR #1554, opened 2016-01-21
Support for bare PGP-encrypted attachments; see issues #1004, #701

JackDca:


JackDca:
Each MIME part with content-disposition "attachment" is sniffed to see if they are PGP objects.  If so, it is unwrapped, then the process is repeated on the contents.  The PGP data is used to restore the original file name - if it is absent, the "filename" field from content-disposition is used, trimming off ".asc" or ".pgp" as appropriate.  Possible enhancments:  There may be performance issues due to the way gpg --list-packets is used to do the sniffing.  Also, it may be better to avoid sniffing some content-types.

BjarniRunar:
Thank you for doing this! Copy-pasting my comments from IRC:

Hmm, could you not use the Python pgpdump instead with in-memory data, to avoid having to invoke gpg on every single attachment? Seems this will slow down everything, all the time, which isn't a good direction for things. I'd only sniff if it's an attachment of  a type we either recognize as encrypted, or don't recognize at all.

If we recognize the mime-type of an attachment as *not* being something encrypted, then we could skip the sniff. :)  This'll take some iteration most likely.

There is some sample code in the plugins/keydiscovery code, we use it to find keys.

JackDca:
To address possible performance issues, in this version GnuPG.sniff() does not use gpg or pgpdump.  It accepts base64, quoted-printable or unencoded data.  Encoded data is decoded in segments only as required.  PGP objects that are armored (before encoding) are recognized from the armor header only.  Non-armored objects are recognized by decoding the PGP packet headers only and checking them for legality.

JackDca:
Most testing of this code has been done using the JackDca/Mailpile bare-pgp-attachments branch which is based on the GitHub Mailpile master code from 2015-10-21.   Additional testing has also been done by merging bare-pgp-attachments into the GitHub Mailpile master code from 2016-02-04.  That merged code is now on JackDca/Mailpile master.

BjarniRunar:
I have merged this into my local branch - am performing a few more tests and cleaning up some other things before I push. Thank you for doing this!

JackDca:
Thank you for your work on Mailpile!  Let me know of any problems or suggestions.  I am ready to do more work on this code if you decide it's needed.

JackDca:
One-line fixes to attachments handling.

The maximum base64 line length per RFC2045 6.8 is 76 characters.  This does NOT include the white space at the end of line.  Discovered this when processing an email from an outlook.com mail box.  Looks like outlook.com modifies base64 encoding by changing the division into lines so the lines are the maximum 76 length. (gpgi.py sniff()).

Catch IndexError exceptions in pgpdump.  A corrupt large attachment (armored with the last part including the -----END PGP ... cut off) caused an IndexError in pgpdump with the result that the incoming email was discarded completely.  Preferably this should be fixed upstream in pgpdump but catching the exception fixes it. (email_keylookup.py _get_keydata() )

PR #1587, opened 2016-04-09
Fix non-spam button

DaLynX:
If action is 'remove', remove the 'spam' tag and add the 'new' tag.


DaLynX:
The notification pop-up stills have to be fixed.

BjarniRunar:
Thank you!

PR #1596, opened 2016-04-24
Fix subtag hiding

### LINKED ISSUE #1595: Cannot hide subtag
I realized subtag were not hidden even when you asked so in the Edit window, which I believe is not expected behaviour.

Fixing this would be in particular useful for automatically created tags matching remote IMAP folders that you might want to hide (_dovecot.sieve_, for instance).

DaLynX:
Fixes #1595


BjarniRunar:
Thanks for working on this!

PR #1598, opened 2016-05-02
Fixes and updates the Dockerfile

aleksandar-todorovic:
Since the current Dockerfile is not really building, I decided to edit it, make it working again and adjusting it so that it uses the newer version of Ubuntu (16.04).

**Worth noting:** This version of the Dockerfile follows the installation instructions for building Mailpile on Linux, so it uses Git to clone the code and then sets things up from there. Could be changed to `ADD` the code from the cloned repository, but in my opinion this is the easier solution since it allows people to just take the Dockerfile out of the repository and build it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
+MAINTAINER Aleksandar Todorović <aleksandar@r3bl.me>
 
 # Install dependencies
 RUN apt-get update -y && \
-    apt-get install -y openssl python-imaging python-jinja2 python-lxml libxml2-dev libxslt1-dev python-pgpdump spambayes
+    apt-get install -y tor python openssl python-imaging git python-jinja2 python-lxml libxml2-dev libxslt1-dev python-pgpdump spambayes python-pip && \
+    apt-get upgrade -y
 
 # Add code
-WORKDIR /Mailpile
-ADD . /Mailpile
+RUN git clone --recursive https://github.com/mailpile/Mailpile/ Mailpile
+WORKDIR "/Mailpile"
 
 # Create users and groups
 RUN groupadd -r mailpile \
@@ -20,16 +22,18 @@ RUN touch /mailpile-data/.gnupg/docker_placeholder
 RUN chown -R mailpile:mailpile /Mailpile
 RUN chown -R mailpile:mailpile /mailpile-data
 
-# Run as non-privileged user
-USER mailpile
+# Install pip requirements
+RUN pip install -r requirements.txt
 
 # Initialize mailpile
 RUN ./mp setup
 
 # Entrypoint
-CMD ./mp --www=0.0.0.0:33411 --wait
 EXPOSE 33411
 
 # Volumes
 VOLUME /mailpile-data/.local/share/Mailpile
 VOLUME /mailpile-data/.gnupg
+
+ENV MAILPILE_HOME=/mailpile-data/.local/share/Mailpile/
+CMD ["./mp", "--www=0.0.0.0:33411", "--wait"]


BjarniRunar:
Could @DoubleMalt take a look and comment on these changes?

Dockerfile
```diff
@@ -1,12 +1,14 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
+MAINTAINER Aleksandar Todorović <aleksandar@r3bl.me>
 
 # Install dependencies
 RUN apt-get update -y && \
-    apt-get install -y openssl python-imaging python-jinja2 python-lxml libxml2-dev libxslt1-dev python-pgpdump spambayes
+    apt-get install -y tor python openssl python-imaging git python-jinja2 python-lxml libxml2-dev libxslt1-dev python-pgpdump spambayes python-pip && \
+    apt-get upgrade -y
```
aviau: > You should avoid RUN apt-get upgrade or dist-upgrade, as many of the “essential” packages from the base images won’t upgrade inside an unprivileged container. 

Source: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

Dockerfile
```diff
@@ -1,12 +1,14 @@
...
 
 # Add code
-WORKDIR /Mailpile
-ADD . /Mailpile
+RUN git clone --recursive https://github.com/mailpile/Mailpile/ Mailpile
```
aviau: Hmm? This dockerfile resides in the repo, why not just mount the code?

Dockerfile
```diff
@@ -1,12 +1,14 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
+MAINTAINER Aleksandar Todorović <aleksandar@r3bl.me>
```
aviau: I wouldn't think this makes sense since this is maintained by the mailpile community

DoubleMalt:
Well ... it's worth updating the Ubuntu version. Other than that the changes seem to do more harm than good. Adding the missing packages to the apt-get install line should be enough.

Dockerfile
aleksandar-todorovic: Didn't I already explain the reason behind it in the description of the pull request?

Dockerfile
aleksandar-todorovic: I agree completely. It's just that I was not able to find someone to contact right away when I noticed that something's wrong with this pull request.

Dockerfile
aleksandar-todorovic: Noted. Thanks for providing me with the link!

aleksandar-todorovic:
@DoubleMalt That comment seems a bit harsh. Mind explaining why? 

I repeat: The whole idea with this pull request is for the Dockerfile to follow the official Ubuntu instructions instead of doing things the other way around.

Dockerfile
aviau: I disagree with your argument. Nobody does that.

Dockerfile
```diff
@@ -20,16 +22,18 @@ RUN touch /mailpile-data/.gnupg/docker_placeholder
 RUN chown -R mailpile:mailpile /Mailpile
 RUN chown -R mailpile:mailpile /mailpile-data
 
-# Run as non-privileged user
-USER mailpile
```
DoubleMalt: Not running as root user is an essential security measure. A process run docker container as the root user can do many things with root rights on the host.

DoubleMalt:
@aleksandar-todorovic Sorry for discouraging you. Most arguments are already brought forward by @aviau, I commented on your removal of the user setting.

Using git to pull in the sources is not a good idea, not only because nobody does that, but because it decouples the Dockerfile from the version on Git. Different Mailpile versions might need different Dockerfiles, so it is best kept in the same directory.

What error did you get building the Mailpile Docker container?

BjarniRunar:
Thanks everyone for chiming in here. and thanks @aleksandar-todorovic for your efforts.

Just to clarify, the reason I asked @DoubleMalt to comment is because he works with CloudFleet who have been trying to build small custom servers using Mailpile, and if I recall correctly, they use the Dockerfile. So I didn't want the changes to break their builds.

If you guys could agree on how things should work and amend the PR to match, I'll be happy to accept it.

Dockerfile
mikefaille: I agree with @aviau . Since we must enforce coerence between Dockerfile and git repository, it's preferable to get all stuff at time. Or, if your really want to "git clone", use a tag or hash please.

JocelynDelalande:
@aleksandar-todorovic bump ?

BjarniRunar:
I think this has timed out; there's a new effort starting up in #1698. Hopefully @mathroc will take a look at what you did and make use of it.

Sorry this didn't get merged, thank you for your efforts!

Issue #1612, opened 2016-06-30
Adding account: what does "download" mean for IMAP and IMAP/TLS?


tuxlifan:
When adding an account, in the "Receiving Mail" section there is the same check-box with label "Download all mail and add to search engine" when selecting IMAP or IMAP/TLS as well as POP3.

_This is really scary and an absolute blocker for me in my "user hat" to continue to try out mailpile:_
- Does it mean that mailpile will download all mail from the IMAP server, **DELETE** it on the IMAP server and then add it to the local search engine?
- If I do not check this box (or uncheck it later) will I not get new mail and will I not be able to use the search engine?

Use case:
I want to synchronize the mail from the IMAP server with the local mailpile storage but **KEEP** the mail on the IMAP server so I can access it also otherwise and then also be able to use "the search engine" in mailpile.

Browsing the source I _think_ that for the non-"dev_version" there is the hidden "source-{{ new_rid }}-leave-on-server" set to yes so the mail should stay on the server, but am I right?
Why not put it out in the open since data (=mail) is the most important thing a MUA is about and the user needs to know what exactly is going on and be in control?

mailpile 1.0.0rc0

Thanks.

PS: The same checkbox also appears for POP3 accounts where normal MUAs usually have separate settings that allow one to set: 1) how many emails or what maximum size to download and 2) if the emails should be deleted on the server after reception. (I don't use POP3 but wanted to point out that also in that case the label is not painting the whole/exact picture and using the identical control for IMAP makes one feel that it might not be safe and trustworthy)

BjarniRunar:
Uhm, I can't think of a single instance where "downloading" from the Internet deletes the original copy by default. That's just not how things work, why would you assume Mailpile is different?

But as I suggested on your pull request, perhaps using the word Copy would be less confusing.

tuxlifan:
My thinking in this instance comes from the good old(?) POP3-days where you downloaded your mail and then the MUA could helpfully delete everything from the server so your mailbox would not overflow after those 20 emails you just received ;)
I agree that the word Copy is a good solution and updated the PR.

BjarniRunar:
I think this is resolved then! :)

PR #1613, opened 2016-06-30
Clarify that "Download all mail" will leave mail on the server

tuxlifan:
possible solution for #1612


BjarniRunar:
How about we just use the word Copy instead of Download? :-)

Making the string 2x as long isn't great for translations or layout, and generally the fewer words the better from a usability pov. Most people don't read good...

JocelynDelalande:
@tuxlifan bump on that ?

JocelynDelalande:
I agree with @BjarniRunar suggestion.

tuxlifan:
Yes, "Copy" sounds like a great solution! (Sorry, been offline for a while.)

BjarniRunar:
Thank you!

PR #1615, opened 2016-07-17
dont select email when switching to html/text view

biji:


BjarniRunar:
Great stuff, thank you.

PR #1618, opened 2016-07-27
Bundled QtWebkit with simple notifications

harindu95:
I have used QWebView and setup automatic login.


BjarniRunar:
Hi!

This is interesting, but you're absolutely right that I cannot merge this as-is; you've disabled all the old functionality and made this a non-optional thing. Not to mention that you've also completely subverted the security of the app by storing the password in a plain-text file.

Thanks for working on this though, I (or someone else) may well be able to build upon your work at a later date.

harindu95:
Hey
yeah I couldn't make head or tails of the python server so I decided to subvert it. If you decide to use qtwebkit in the future, it seems to crash possibly due to incorrect html in some emails.
Thanks for the reply

PR #1621, opened 2016-09-02
Use GPGME

flanfly:
This replaces the GnuPG interface with GPGME to allow Mailpile to work with GnuPG 2.1 and 2.0.

### DIFF
--- a/mailpile/config/manager.py
+++ b/mailpile/config/manager.py
@@ -281,21 +281,19 @@ class ConfigManager(ConfigDict):
         return rv
 
     def load_master_key(self, passphrase, _raise=None):
-        keydata = []
+        keydata = None
 
         if passphrase.is_set():
-            parser = lambda d: keydata.extend(d)
             try:
                 with open(self.conf_key, 'rb') as fd:
-                    decrypt_and_parse_lines(fd, parser, self,
-                                            newlines=True,
-                                            passphrase=passphrase)
+                    gpg = GnuPG(None)
+                    keydata = gpg.decrypt(fd.read(),passphrase=passphrase.get_reader())[2]
             except IOError:
-                keydata = []
+                pass
 
         if keydata:
             self.passphrases['DEFAULT'].copy(passphrase)
-            self.master_key = ''.join(keydata)
+            self.master_key = keydata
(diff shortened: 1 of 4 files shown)


BjarniRunar:
Hello Kai!

This is amazing work, thank you. However, if I merge this as-is, it'll break e-mail for all our current testers/users...

Can we simply move the gpgme based code to a separate module, leaving the 1.4.x code intact, and choose between them at runtime? At least during a transition period.

nwalfield:
Hi Bjarni!

gpgme supprts not only 2.0 & 2.1, but also 1.4.  If anything doesn't work with 1.4, then that is a bug (minus features that were only added to 2.0 or 2.1, of course).  So, I think Kai meant to say "to allow Mailpile to *also* work with GnuPG 2.1 and 2.0."

:) Neal

BjarniRunar:
OK, good to know.

This is still an all-or-nothing change, and it's completely unclear to me at this point whether/how all the edge cases we'd worked around in the old code are handled by the new, and how this code will behave with the gpg and gpgme that come with not-quite-brand-new Linux distros.

Since we do have a few very precious users at the moment who are used to being able to git pull from HEAD now and then, I'd like to make sure this doesn't break their mail and I'd like to "canary" it a bit before chucking the old code in the bit bucket.

However, considering the amount of work you've already put in, I should probably do that myself. I'll see about merging locally and make the tweaks I need.

nwalfield:
I can apprecaite the desire to not want to break things for your existing users.

If there are any functional regressions with the current code, please don't hesitate to raise the issues (either here or, for quicker feedback, on the jabber channel or the gnupg-devel mailing list) and we'll try and help.

BjarniRunar:
This is not dead! I have started working on a merge in my local tree.

BjarniRunar:
After thinking about this a great deal, I am not actually in favour of switching to GPGME.

I have opened a discussion issue, #1742 to capture arguments pro and con.

I have archived this code in my local repository; I am going to close this PR for now. It cannot be merged in it's current state - conflicts aside, this would break any existing Mailpile installations. If #1742 leads to a decision to make this change, I'll base the port on these commits so Kai still gets credit for his work.

PR #1625, opened 2016-09-26
Improve mail source tag creation and policies

BjarniRunar:
This PR improves and simplifies how Mailpile mail sources create tags.

1. The slugs/links now have meaning (instead of just being unique IDs)
2. Fix the "Organize" sidebar mode to give direct access to tag settings and show archived tags
3. Simplify tag creation policy: always add to sidebar, never create labels

Individual commit messages have more details.

The sidebar org mode change picks up a few minor changes to the default less-compiled CSS as a side-effect.

### DIFF
--- a/mailpile/plugins/contacts.py
+++ b/mailpile/plugins/contacts.py
@@ -1161,7 +1161,9 @@ class AddProfile(ProfileVCard(AddVCard)):
                     'label': False,
                     'display': 'invisible'
                 })
-                tags[vcard.tag].slug = 'account-%d' % len(tags)
+                from mailpile.plugins.tags import Slugify
+                tags[vcard.tag].slug = Slugify(
+                    'account-%s' % vcard.email, tags=self.session.config.tags)
 
         route_id = state.get('route_id')
         if route_id:
--- a/mailpile/plugins/contacts.py
+++ b/mailpile/plugins/contacts.py
@@ -971,7 +971,6 @@ def ProfileVCard(parent):
                     # so we save config to trigger instanciation.
                     self._background_save(config=True, wait=True)
                     src_obj = config.mail_sources[src_id]
-                    src_obj.set_tag_visibility(self._yn(prefix + 'visible-tags'))
 
                 else:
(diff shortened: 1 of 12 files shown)


BjarniRunar:
Ping @JocelynDelalande - shall we attempt a review? :-)

mailpile/plugins/contacts.py
```diff
@@ -1161,7 +1161,9 @@ def _update_vcard_from_post(self, vcard, state=None):
                     'label': False,
                     'display': 'invisible'
                 })
-                tags[vcard.tag].slug = 'account-%d' % len(tags)
+                from mailpile.plugins.tags import Slugify
+                tags[vcard.tag].slug = Slugify(
+                    'account-%s' % vcard.email, tags=self.session.config.tags)
```
karlp: I guess email is the best you've got, no guarantee of having a contact name in the vcard I suppose?

BjarniRunar: Yes. Also, this is a profile or account - so the name would be your own name most of the time. Not a useful differentiator in the common case.

shared-data/default-theme/css/default.css
```diff
@@ -460,7 +460,8 @@
 .icon-mime[type="application/x-tar"]:before,
 .icon-mime[type="application/x-winzip"]:before,
 .icon-mime[type="application/x-zip"]:before,
-.icon-mime[type="application/x-zip-compressed"]:before {
+.icon-mime[type="application/x-zip-compressed"]:before,
+.icon-mime[type="application/x-rar-compressed"]:before {
   content: "\e66b";
```
BjarniRunar: Yes. I did mention it in the PR comment. :)

BjarniRunar:
Regarding subtags vanishing from the sidebar - this is required at the moment, since the drag-drop used to reorder things in org mode can't handle the subtags sensibly. Much more code/thought is needed to address that, so for now we just make it clear that org mode can't handle it by making them disappear.

PR #1648, opened 2016-10-06
Bug-fixes and improvements to mail source UI

BjarniRunar:
This PR makes the following changes, in order of decreasing importance:

   * Fix hard-coded limits which would cause Mailpile to stop discovering new mail/boxes
   * Fix UI notifications behaviour for mail-source related events
   * Make it possible to rename mail sources (by using the UI to rename the main mail source tag)
   * Change the color of the Add Account button, to make it more obvious something has changed after reviewing the privacy policy

### DIFF
--- a/mailpile/config/defaults.py
+++ b/mailpile/config/defaults.py
@@ -181,6 +181,7 @@ CONFIG_RULES = {
             'visible_tags':(_('Make tags visible by default?'), bool, True),
             'process_new': (_('Is a potential source of new mail'), bool, True),
             'apply_tags':  (_('Tags applied to messages'), str, []),
+            'max_mailboxes':(_('Max mailboxes to add'), int, 100),
         }),
         'mailbox': (_('Mailboxes'), {
             'name':        (_('The name of this mailbox'), str, ''),
--- a/mailpile/mail_source/__init__.py
+++ b/mailpile/mail_source/__init__.py
@@ -31,8 +31,7 @@ class BaseMailSource(threading.Thread):
     SAVE_STATE_INTERVAL = 3600  # How frequently we pickle our state
     INTERNAL_ERROR_SLEEP = 900  # Pause time on error, in seconds
     RESCAN_BATCH_SIZE = 200     # Index at most this many new e-mails at once
-    MAX_MAILBOXES = 100         # Max number of mailboxes we add
-    MAX_PATHS = 5000            # Abort if asked to scan too many directories
+    MAX_PATHS = 2000            # Limit how many directories we scan at once
 
     # This is a helper for the events.
     __classname__ = 'mailpile.mail_source.BaseMailSource'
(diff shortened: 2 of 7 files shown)


BjarniRunar:
Leaving here for a few days before merging, in case anyone (@JocelynDelalande ?) wants to review it.

JocelynDelalande:
@BjarniRunar ok, I'm giving a look.

mailpile/mail_source/__init__.py
```diff
@@ -163,6 +162,8 @@ def _save_config(self):
             self.session, 'Save config', self.session.config.save)
 
     def _log_status(self, message):
+        # If the user renames our parent_tag, we assume the name change too.
+        self.update_name_to_match_tag()
```
JocelynDelalande: Is it really the good place to call something that seems unrelated to logging ?

mailpile/config/defaults.py
```diff
@@ -181,6 +181,7 @@
             'visible_tags':(_('Make tags visible by default?'), bool, True),
             'process_new': (_('Is a potential source of new mail'), bool, True),
             'apply_tags':  (_('Tags applied to messages'), str, []),
+            'max_mailboxes':(_('Max mailboxes to add'), int, 100),
```
JocelynDelalande: That's  an IMAP-only parameter, maybe the var name should reflect that.

mailpile/mail_source/__init__.py
```diff
@@ -31,8 +31,7 @@ class BaseMailSource(threading.Thread):
     SAVE_STATE_INTERVAL = 3600  # How frequently we pickle our state
     INTERNAL_ERROR_SLEEP = 900  # Pause time on error, in seconds
     RESCAN_BATCH_SIZE = 200     # Index at most this many new e-mails at once
-    MAX_MAILBOXES = 100         # Max number of mailboxes we add
-    MAX_PATHS = 5000            # Abort if asked to scan too many directories
+    MAX_PATHS = 2000            # Limit how many directories we scan at once
```
JocelynDelalande: same here (about naming)

shared-data/default-theme/html/jsapi/ui/notifications.js
```diff
@@ -38,6 +40,7 @@ Mailpile.cancel_notification = function(not_id, $existing, replace, record) {
 
 
 Mailpile.notification = function(result) {
+  Mailpile.expire_canceled_notifictions();
```
JocelynDelalande: Good time to fix that typo : `notifictions` -> `notifications`

BjarniRunar: Good catch! Thanks, will fix.

mailpile/config/defaults.py
JocelynDelalande: ok

mailpile/mail_source/__init__.py
JocelynDelalande: > Well, the logging is what triggers UI updates (events, notifications) so ensuring we have the right idea about what our name is before doing so is reasonable. 

> But generally speaking, this is pure side-effect code and as such is ugly almost anywhere we put it. It's a hack, but without triggers on the settings themselves, it's hard to do this cleanly.

We do agree on the fact that side effects are bad. But given we have a side-effect method, let's not hide its call it in a method (`log_status`) that doesn't give a clue about the fact it will trigger side effect on the object itself.

Issue #1651, opened 2016-10-19
"Got It" button doesn't have clickable cursor


Epse:
The little box that pops up for example upon first viewing a thread underneath the people has a "Got it" button, which unfortunately has a text selection (default) cursor. Setting `cursor:pointer;` on the appropriate class should fix it.

Unfortunately now I closed it and I cant inspect to see the class, so if anyone knew that I would fix it.

Epse:
Fixed by pull request

Epse:
accidentially closed

Epse:
now fixed by pull request referenced above

PR #1653, opened 2016-10-19
fixed #1561 Got It button has wrong cursor

### LINKED ISSUE #1561: BCC of encrypted messages should send/encrypt separately
The user expectation when using BCC is that the people in the To/CC lines are unaware of the BCC'ed recipients.

PGP may violate this expectation, by including the key IDs of the BCC'ed folks. We should fix this by making sure BCC'ed folks get a separate copy of the message that is encrypted only to them.

Relates to #733.

Epse:
The "Got it" button had a text select cursor, as described in the issue. Only the checkmark icon had a pointer, now the clickable text has the correct cursor as well.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -3348,6 +3348,10 @@ input[type="button"].button-small {
   user-select: none;
   transition-duration: 0.2s;
 }
+/* OK, Got it */
+.ok-got-it {
+  cursor: pointer;
+}
 /* Primary */
 button.button-primary,
 input[type="submit"].button-primary,


NoahAndrews:
I think you mean #1651.

BjarniRunar:
Note: I've merged this, but I will have to fix it at my end - you edited default.css directly, which is an autogenerated file. In the future please also update the less source files! :-)

BjarniRunar:
I've fixed the less problem in my source tree, so don't worry about it. It'll be in my next big push.

Epse:
Ah well, I could not find certain classes in the less, but guess I'll have to look better

PR #1654, opened 2016-10-20
fixed #1627 - added tooltip to compose button on profile page

### LINKED ISSUE #1627: (trivial) compose icon on home page has no tooltip
On the "home" page, the green "torso" icon has the tooltip "Edit account" but the compose icon beside has none.  Something like, "Compose as this sender..."?  related to https://github.com/mailpile/Mailpile/issues/1306 but trivial frontend work...

Epse:
As said, fixed #1627 

I added a tooltip, namely "Compose as example@example.com" where obviously the selected address replaces the example.

I chose the approach with the address since people tend to have the same name on different accounts and it is the address that counts.

### DIFF
--- a/shared-data/default-theme/html/profiles/index.html
+++ b/shared-data/default-theme/html/profiles/index.html
@@ -94,7 +94,7 @@
     <tr valign="top" class="message">
        <td width='1%'>
          {%- if route %}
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Thanks for doing this!

PR #1659, opened 2016-10-25
small UI improvements

Futal:
Broken link, drop-down menu refresh, string to be translated


BjarniRunar:
Thanks for doing this! I'm a little baffled that the relative URL doesn't work and replacing it with an absolute one does... but I believe you. ;-)

Futal:
The URL problem seems to come from a bug in `urlmap.py`. Both Firefox and Chrome show the full correct URL in the status bar. If copied-pasted, the URL works. Yet when clicking the link, it triggers some incorrect URL rewriting.

PR #1666, opened 2016-11-01
Security: Disable most local encryption by default

BjarniRunar:
In this PR:

   * Change defaults for local data encryption settings to False
   * Hide the local-encryption settings from the privacy page

I may end up reverting this later, but the rationale for this is
as follows:

   1. Users with strong security needs should encrypt their HDDs
   2. Encryption of local data slows things down
   3. Encryption of local data increases odds of data loss & bugs
   4. Until our local encryption is faster and more robust (forking
      openssl does not work very well), concerns 2 and 3 outweigh
      the security concern for users not covered by point 1.

Further, our current implementation relies on "openssl enc", which
was found by some researchers to be disappointingly weak earlier
this year:

   https://cryptosense.com/weak-key-derivation-in-openssl/

Thanks to Kai Michaelis for the heads' up on the openssl issue.


JackDca:
I use Mailpile for all my email needs with local encryption enabled.  I'm not seeing speed issues.  I have not had data loss problems.  Maybe this is a problem if you are running on a Raspberry Pi but ... .

At least leave me the capability of turning encryption on in the UI.

Just my opinion.

BjarniRunar:
Well, the point is to discourage users from using the feature.

While it's of questionable utility and known to have poor security, encouraging users to switch it on is just asking for trouble. The feature is still available from the CLI; users who know what they are doing can switch it on there.

It's a tough call; data integrity vs. data privacy in the event that a device is stolen. Note that the settings and metadata index remain encrypted, and the search index is still obfuscated (hashed), because the trade-offs are different there.

JackDca:
Hello Bjarni Runar,

I get your point, and I certainly accept that the default can reasonably be switched to "no local encryption".  What I don't like is hiding the local encryption option from the casual user.  In previous situations that you have blogged about, you have pointed out that weak security is better than no security IF it does not lead to false confidence.  For me, the promise of Mailpile is that it will let "casual users" who don't "know what they are doing" use encryption, moving us toward a world where the use of end-to-end encryption is the norm.  So when I see a feature being disabled in the user interface, that concerns me, because a casual user is not likely to use the CLI.  (Even I find it painful to use sometimes.)

Casual users may not know how to turn on full disk encryption on an existing system, so that suggestion isn't much help to them..  Also, the easiest way to do backup right now is to copy the entire Mailpile data directory to a USB drive or to the cloud.  If that data is not encrypted, the backup copy becomes a security weakness.

You have earned the right to make the call - but that is my opinion.

BjarniRunar:
Thank you for your thoughts, @JackDca, I really appreciate them. And your point about non-encrypted backups being a weakness is very good, I hadn't really given that much thought. So thanks.

As you can see, I haven't merged this yet... I'm going to try and take a look over the next few days at how hard it will be to just fix the local encryption so this becomes less necessary.

JackDca:
Thanks, as always, for all your work ... and for having an open mind!

BjarniRunar:
I am closing this; instead of disabling the local encryption I am adding a dependency to the project (add a library so we can stop using `openssl enc`) and switching encryption algorithms to AES in CTR mode, which doesn't destroy all the data just because single bit got flipped.

This addresses the above concerns and lets us keep the encryption-by-default. It's mostly written, a PR will appear soon.

BjarniRunar:
See #1684. :-)  Thanks for the discussion @JackDca.

PR #1667, opened 2016-11-01
WIP: Many, many bug-fixes to allow browsing mailboxes and sources

BjarniRunar:
This PR is big; it includes a major reorganization of how the search index is used, to allow for multiple indexes. The multiple index feature is then used to make it possible to "search" inside a mailbox that hasn't actually been added to Mailpile. That is then used by the browser interface.

There are also a bunch of bugfixes here to do with all sorts of sources and searching.

I'm still working on this, and may add more commits over the next few days, but wanted to push this so people could see what I am doing and maybe give some feedback.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,15 +78,16 @@ web: less js
 
 alltests: clean pytests
 	@chmod go-rwx mailpile/tests/data/gpg-keyring
-	@DISPLAY= python2 scripts/mailpile-test.py || true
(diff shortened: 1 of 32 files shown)


BjarniRunar:
OK, review window has timed out. Merging this, I've had it running locally for quite some time and it definitely fixes more things than it breaks. ;-)

Issue #814, opened 2014-08-09
Implement support for Libravatar and/or X-Face


bnvk:
In addition to getting contact pictures from Gravatar, we should also support getting pics from https://www.libravatar.org , the X-Face header, and other alternate sources.

ryneeverett:
I looked at several other email clients that support x-face and they all use compface and netpbm to convert it into an image. I doubt we want to add those as mailpile dependencies, but, given the low resolution of x-face images, I would think it would be feasible to convert into a bitmap format that can be displayed in html (bmp?) without too much of a performance cost.

PR #1669, opened 2016-11-03
Add "Face" header vcard, per #814.

ryneeverett:
Tested successfully with claws-mail.

### DIFF
--- a/mailpile/plugins/setup_magic.py
+++ b/mailpile/plugins/setup_magic.py
@@ -283,9 +283,6 @@ class SetupMagic(Command):
         if not vcard_importers.libravatar:
             vcard_importers.libravatar.append({'active': True})
             session.ui.notify(_('Enabling Libravatar image importer'))
-        if not vcard_importers.face:
-            vcard_importers.face.append({'active': True})
-            session.ui.notify(_('Enabling Face image importer'))
 
         gpg_home = os.path.expanduser('~/.gnupg')
         if os.path.exists(gpg_home) and not vcard_importers.gpg:
--- a/mailpile/plugins/setup_magic.py
+++ b/mailpile/plugins/setup_magic.py
@@ -283,6 +283,9 @@ class SetupMagic(Command):
         if not vcard_importers.libravatar:
             vcard_importers.libravatar.append({'active': True})
             session.ui.notify(_('Enabling Libravatar image importer'))
+        if not vcard_importers.face:
+            vcard_importers.face.append({'active': True})
+            session.ui.notify(_('Enabling Face image importer'))
 
         gpg_home = os.path.expanduser('~/.gnupg')
         if os.path.exists(gpg_home) and not vcard_importers.gpg:
--- a/mailpile/plugins/vcard_face.py
+++ b/mailpile/plugins/vcard_face.py
@@ -1,4 +1,6 @@
+import sys
 import time
+import warnings
 
 from mailpile.i18n import gettext as _
 from mailpile.mailutils import Email
@@ -27,9 +29,11 @@ class FaceImporter(VCardImporter):
         results = []
 
         for vcard in self.session.config.vcards.values():
+            email = vcard.email
+
             # Get latest message from vcard email address.
             from_messages = list(mail_index.search(
-                self.session, ['from:' + vcard.email]).as_set())
+                self.session, ['from:' + email]).as_set())
             if not from_messages:
                 continue
             mail_index.sort_results(self.session, from_messages, 'date')
@@ -38,7 +42,12 @@ class FaceImporter(VCardImporter):
             # Add face vcard.
             face = mail_index.hdr(latest_message, 'face')
             if face:
-                results.append(self._make_vcard(vcard.email, VCardLine(
+                if sys.getsizeof(face) > 998:
+                    warnings.warn(
+                        email + "'s face header exceeds the maximum size. "
+                        "See the spec: http://quimby.gnus.org/circus/face/")
+                    continue
+                results.append(self._make_vcard(email, VCardLine(
                     name='photo',
                     value='data:image/png;base64,%s' % face,
                     media_type='image/png')))
--- a/mailpile/plugins/vcard_face.py
+++ b/mailpile/plugins/vcard_face.py
@@ -3,7 +3,6 @@ import time
 from mailpile.i18n import gettext as _
 from mailpile.mailutils import Email
 from mailpile.plugins import PluginManager
-from mailpile.search import MailIndex
 from mailpile.vcard import MailpileVCard, VCardImporter, VCardLine
 
 
@@ -24,8 +23,7 @@ class FaceImporter(VCardImporter):
     VCARD_IMG = ''
 
     def get_vcards(self):
-        mail_index = MailIndex(self.session.config)
-        mail_index.load(self.session)
+        mail_index = self.session.config.get_index(self.session)
         results = []
 
         for vcard in self.session.config.vcards.values():
--- a/mailpile/plugins/vcard_face.py
+++ b/mailpile/plugins/vcard_face.py
@@ -1,0 +1,57 @@
+import time
+
+from mailpile.i18n import gettext as _
+from mailpile.mailutils import Email
+from mailpile.plugins import PluginManager
+from mailpile.search import MailIndex
+from mailpile.vcard import MailpileVCard, VCardImporter, VCardLine
+
+
+_plugins = PluginManager(builtin=__file__)
+
+
+class FaceImporter(VCardImporter):
+    """
+    Get avatar from Face header.
+    """
+    FORMAT_NAME = 'Face'
+    FORMAT_DESCRIPTION = _('Get avatar from Face header.')
+    SHORT_NAME = 'face'
+    CONFIG_RULES = {
+        'active': [_('Enable this importer'), bool, True],
+    }
+    VCARD_TS = 'x-face-ts'
+    VCARD_IMG = ''
+
+    def get_vcards(self):
+        mail_index = MailIndex(self.session.config)
+        mail_index.load(self.session)
+        results = []
+
+        for vcard in self.session.config.vcards.values():
+            # Get latest message from vcard email address.
+            from_messages = list(mail_index.search(
+                self.session, ['from:' + vcard.email]).as_set())
+            if not from_messages:
+                continue
+            mail_index.sort_results(self.session, from_messages, 'date')
+            latest_message = Email(mail_index, from_messages.pop()).get_msg()
+
+            # Add face vcard.
+            face = mail_index.hdr(latest_message, 'face')
+            if face:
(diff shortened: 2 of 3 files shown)


BjarniRunar:
Very cool!

I have some minor quibbles with the code, I'll add comments inline using the github review thingo.

mailpile/plugins/vcard_face.py
```diff
@@ -0,0 +1,57 @@
...
+    FORMAT_DESCRIPTION = _('Get avatar from Face header.')
+    SHORT_NAME = 'face'
+    CONFIG_RULES = {
+        'active': [_('Enable this importer'), bool, True],
+    }
+    VCARD_TS = 'x-face-ts'
+    VCARD_IMG = ''
+
+    def get_vcards(self):
+        mail_index = MailIndex(self.session.config)
```
BjarniRunar: This is not the right way to get a mail index object - you've just created an entire new index and loaded from disk. For a user with a lot of mail, this will take a very long time and waste a lot of computing time. Instead:

    mail_index = self.session.config.get_index(self.session)

mailpile/plugins/vcard_face.py
```diff
@@ -0,0 +1,57 @@
...
+        results = []
+
+        for vcard in self.session.config.vcards.values():
+            # Get latest message from vcard email address.
+            from_messages = list(mail_index.search(
+                self.session, ['from:' + vcard.email]).as_set())
+            if not from_messages:
+                continue
+            mail_index.sort_results(self.session, from_messages, 'date')
+            latest_message = Email(mail_index, from_messages.pop()).get_msg()
```
BjarniRunar: This is an expensive operation. Can we avoid doing this for every single contact, every time?

A couple of ideas:

1. make the search more exact, so it doesn't return any results if messages from that user don't have a Face header.

2. don't perform the search every time if the user already has an avatar (maybe use a random number and only search 1/1000 times, or see if there is a timestamp somewhere we can check and use)

Ideally, the code would do both.

mailpile/plugins/vcard_face.py
```diff
@@ -0,0 +1,57 @@
...
+
+            # Add face vcard.
+            face = mail_index.hdr(latest_message, 'face')
+            if face:
+                results.append(self._make_vcard(vcard.email, VCardLine(
+                    name='photo',
+                    value='data:image/png;base64,%s' % face,
```
BjarniRunar: Can we add some sort of sanity check to prevent the most obvious attacks through this vector? Consider what sort of mischief people could get up to by sending e-mails with malicious Face: headers...

Possible attacks:

1. very large images, bloating the vcard
2. malformed content of some sort

mailpile/plugins/setup_magic.py
```diff
@@ -283,6 +283,9 @@ def basic_app_config(self, session,
         if not vcard_importers.libravatar:
             vcard_importers.libravatar.append({'active': True})
             session.ui.notify(_('Enabling Libravatar image importer'))
+        if not vcard_importers.face:
+            vcard_importers.face.append({'active': True})
+            session.ui.notify(_('Enabling Face image importer'))
```
BjarniRunar: Until this feature has been tested a bit, I am not sure it should be enabled by default.  Especially see my concerns below about security... :-)

BjarniRunar:
More thoughts... (this is an exciting PR, so it's got me thinking :-)

Rather than adding a VCardImporter, did you consider adding a hook which looks for Face: header as incoming mail is scanned? That approach would avoid most of the inefficiencies of the above code, since each message would be checked for a Face: header exactly once and you would be working with data that had already been parsed for the search indexer anyway.

My thoughts on these things have been evolving over time, but currently the way I'd recommend adding new features like this is as follows:

1. Create an API command which scans a set of messages for Face: headers, adding them to the appropriate VCards if necessary. This facilitates two things: manual testing/experimentation, and UI support. See the `mailpile/plugins/exporters.py` for a smallish example.

2. Make the core logic from 1. shared, so it can be invoked from elsewhere.

3. Hook into the message indexing flow when you want to do things with every incoming message. Add search terms to makes messages findable, add tags if you want per-messages UI features. In particular, doing things this way might allow us to display Face: avatars without adding the user to the contact list first.

4. Create a VCardImporter if there is a single external source of VCard data (this doesn't really seem to be the case here).

... generally, Mailpile currently has too much magic, too many things that happen behind the scenes without allowing the user to have a say in things. Note I think you did a great job emulating what we had already done, not your fault I regret some of the design decisions we made! If reorganizing this is too much work, then I'm happy to accept the PR once you've addressed the review and we this will just be added to the laundry list of things to improve later.

mailpile/plugins/vcard_face.py
ryneeverett: > make the search more exact, so it doesn't return any results if messages from that user don't have a Face header.

This would be nice regardless of other performance optimizations, but I didn't find an existing way to do this. Would this require adding search infrastructure?

> don't perform the search every time if the user already has an avatar (maybe use a random number and only search 1/1000 times, or see if there is a timestamp somewhere we can check and use)

It seems like the expense of this operation compared with other vcard lookups that make remote calls is relative to the machine -- is it constrained for local resources (cpu/memory) or bandwidth? Wouldn't the same caching optimizations would be relevant to all of them?

mailpile/plugins/vcard_face.py
ryneeverett: I addressed point **1**, which was straightforward since the spec has a very small size allowance.

I'm not sure about point **2**. Any string fitting the size requirement is valid. Regarding the attack vector, isn't there some existing mechanism to sanitize user input? I've been doing mostly django lately so I'm accustomed to the template system taking care of this.

ryneeverett:
*Some relevant discussion is in the diffs that github considers outdated.*

> Rather than adding a VCardImporter, did you consider adding a hook which looks for Face: header as incoming mail is scanned?

I started to take this approach at first but changed my mind for a few reasons:

1. Consistency with existing vcard importers. Any future implementations which might use the email as an avatar source can be inserted into this importer's loop, so it's consistent with all future importers as well.
2. This is a non-critical task and should be executed asynchronously. It seems like the cost of the email lookup can be reduced to the point where it's not worth adding yet another thread when there's already a mechanism for asynchronously syncing vcards.

> each message would be checked for a Face: header exactly once

It isn't entirely clear to me that that is ideal. Maybe once-per-email is more than necessary. If a contact doesn't consistently set the header, maybe it's acceptable that the behaviour is undefined? In fact, the behaviour is already undefined since it's based on whatever order the importers happened to have run in.

I may hold off on the design-level issues until we've hammered out the implementation details and behaviour.

mailpile/plugins/vcard_face.py
BjarniRunar: You probably want to add a "has:face" keyword, by registering a hook with register_meta_kw_extractor(), see `mailpile/plugins/cryptostate.py` for examples of that.

Basically you'll just check if the parsed message has a Face header and add return `set(["face:has"'])` (yes, backwards, sorry internals are weird) if so - probably about four lines of code. You can then efficiently search for "has:face".

Regarding performance - the importers are invoked periodically and will attempt to synchronize ALL contacts. So if you have 10000 contacts, your current code will attempt to load, parse and evaluate 10000 e-mails. That is a lot of disk seeks and CPU time to accomplish nothing much in the common case where nothing has changed. And no, there's no caches that will help - but this will happily flush any caches we were relying on elsewhere, thus magnifying the performance hit even further. This goes down if we scope things and search for has:face - until everyone starts adding Face: headers to their mail. :-)

mailpile/plugins/vcard_face.py
BjarniRunar: 1. Great!

2. This is hard. Ideally there would be something that validates that it is a valid PNG, but I don't know code that will do this. Solving this sort of thing is very tricky, since if we auto-invoke some C library that parses PNGs to check the validity, that library may then have flaws and our sanity checker gets exploited... not validating at all and just passing through may in fact be safer, since we can probably assume that the browser's PNG parser is pretty well tested. What do you think?

BjarniRunar:
Fair enough. As I said, you did a good job being consistent. Thanks for that!:-)

But regarding the non-criticality of this, it is super cheap to do things at indexing time so adding code that only runs if a Face: header is present, and only works with data that has already been parsed and is ready in RAM...

Regarding the cost of the e-mail lookup being reduced; I don't agree with your optimism there. Opening a mailbox, loading an e-mail and parsing the e-mail... these are and will for the forseeable future be the most expensive operations in Mailpile.

That said - you've written the Importer, so don't go deleting it. The best solution may be to improve the shared importer API so importers can also do work at mail scanning time when appropriate. So let's just try to make this as efficient as possible.

BjarniRunar:
(also, see more comments in the "outdated" review above)

mailpile/plugins/vcard_face.py
ryneeverett: I don't think this issue is particular to this importer, even though it's more likely to be an invalid image than those from other sources.

We don't need to verify that it's a valid png for security reasons, we just need to escape/sanitize it at some point between parsing and rendering to the client to avoid XSS attacks. This should really happen to all photo's anyway in case one of the avatar services is compromised.

Hopefully that is already taken care of and you're just talking about the user experience issue when a broken image is displayed. The image selection algorithm (currently just the latest imported) needs to get smarter anyway, and I would think "broken image automatically reported by client javascript" would be just another (heavily weighted) factor in the ranking.

ryneeverett:
Fair enough. I dismissed your initial comment as high-level and irrelevant to the implementation details before, but I now see that by improving the API I can implement this feature without sacrificing elegance (consistency with the plugin API) or performance. The current plugin API is incomplete which is why my implementation using it is inefficient.

mailpile/plugins/vcard_face.py
BjarniRunar: You're right it's probably not specific to this importer.

It's more serious with this importer though, because sending a forged e-mail from someone plausibly in your contact list, with a nasty Face: header, is a much easier attack than hijacking and corrupting the contents of your contacts' gravatar accounts (for example).

An exploitable bug in this code becomes a worm. An exploitable bug in the other importers does not.

An no, it's not already taken care of.

In general it's NOT safe to assume that "security stuff" is happening elsewhere in Mailpile. There is no mature framework for writing e-mail clients in Python - maybe Mailpile will evolve and become just that at some point, but until then we live in a very different universe from web apps. :-)

BjarniRunar:
Just to clarify - I didn't mean to suggest you should fix the Importer API. You're welcome to work on that too if you like, but that should be a different PR.

BjarniRunar:
And also - yes, all the plugin APIs are incomplete. Quite frankly, we didn't understand the problem space well enough to build good APIs, and this back-and-forth here is part of gaining that experience. This is a very useful discussion IMO.

BjarniRunar:
Does #1681 make this obsolete? Can we close this?

ryneeverett:
Yep.

Issue #1665, opened 2016-11-01
clear search field - icon / button missing


ferdiga:
1.0.0rc

currently it is not possible to clear the search field using the mouse (or the finger on a touch screen device)

suggestion: add a "clear" icon like in thunderbird
![image](https://cloud.githubusercontent.com/assets/6248560/19882406/cf0e8cbe-a00d-11e6-9c4e-9e2330a0af35.png)

BjarniRunar:
Just to play devil's advocate: why does this matter? Even if you've cleared the search field you're going to need to use a keyboard to type in the search terms anyway...

To answer my own question: I guess many users don't know how to select-all and then delete with hot-keys?

ferdiga:
well I thought working with a mouse device should be supported - it may have a reason why a lot of UI have this feature.

shreab373:
@ferdiga hello! i'm a complete noob to open source contribution, do you think i should take this up? could you guide me? :)

ferdiga:
sorry can't give advice - not my domain

BjarniRunar:
@shreab373 Welcome!

I think this task has a few parts, some harder than others:

1) You'd want to find or create an appropriate icon. At the simplest, this would be a Unicode character that is available in the standard fonts, medium complexity would be to create a .png icon and add to the project's static resources (`shared-data/default-theme/img/`), most complex would be to draw a new icon and add to our icon font.

2) Update the HTML/CSS to display the icon. The Mailpile HTML/CSS code lives in `shared-data/default-theme/...`

3) Create a small amount of Javascript which clears the search box when the icon is clicked. The main app Javascript is in `shared-data/default-theme/html/jsapi/...`

Does this help? You may also want to join our IRC channel on Freenode (#mailpile) to have live chats with us there.

PR #1678, opened 2016-11-13
Prevent unwanted selection in search field

kehugter:
The whole search query is selected only if the search field didn't have
focus before the click event. This allows the user to select substrings
and cursor positions using the mouse.


BjarniRunar:
Thanks for doing this! I kinda have the feeling that this whole auto-select thing is a misfeature, but you've at least made it behave better. ;-)

kehugter:
I'm not really a fan of auto-selecting. The only useful situation I can think of is to delete the whole text, but in that case #1665 is IMO more predictable/convenient.

Note that Google doesn't auto-select in any of its products.

PR #1680, opened 2016-11-15
Sanitize avatars to protect against XSS attacks.

ryneeverett:
Avatars are sourced from third parties we have no particular reason to
trust.

This is a followup to discussion in #1669.


BjarniRunar:
Very nice, thank you!

PR #1681, opened 2016-11-15
Plugin and vcard improvements.

ryneeverett:
These changes may seem unrelated, but they all have the goal of implementing the face vcard plugin efficiently by improving the flexibility of the plugin system rather than adding yet another special-purpose plugin type or hook. My premise is that the coupling of plugin typing/configuration *(what/how)* with the hook *(when)* is unnecessary.

This is a followup to discussion in #1669.

### DIFF
--- a/mailpile/plugins/__init__.py
+++ b/mailpile/plugins/__init__.py
@@ -778,15 +778,6 @@ class PluginManager(object):
         return [elem for elem in self.UI_ELEMENTS[ui_type]
                 if context in elem['context']]
 
-    ##[ Pluggable Hooks ]####################################################
-
-    INCOMING_EMAIL = {}
-
-    @staticmethod
-    def trigger(hook, **kwargs):
-        for func in hook.values():
-            func(**kwargs)
-
 
 ##[ Backwards compatibility ]################################################
 
--- a/mailpile/plugins/__init__.py
+++ b/mailpile/plugins/__init__.py
@@ -25,9 +25,9 @@ __all__ = [
     'eventlog', 'search', 'tags', 'contacts', 'compose', 'groups',
     'dates', 'sizes', 'autotag', 'cryptostate', 'crypto_gnupg',
     'setup_magic', 'exporters', 'plugins', 'motd',
-    'vcard_carddav', 'vcard_gnupg', 'vcard_gravatar', 'vcard_libravatar',
-    'vcard_mork', 'html_magic', 'migrate', 'smtp_server', 'crypto_policy',
-    'keylookup'
+    'vcard_carddav', 'vcard_face', 'vcard_gnupg', 'vcard_gravatar',
+    'vcard_libravatar', 'vcard_mork', 'html_magic', 'migrate', 'smtp_server',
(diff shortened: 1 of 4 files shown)


BjarniRunar:
Thanks for doing this! It's going to take me some time to review this properly, but I'll try to get to it as soon as possible.

mailpile/plugins/__init__.py
```diff
@@ -635,16 +635,11 @@ def register_filter_hook_post(self, name, hook):
     VCARD_EXPORTERS = {}
     VCARD_CONTEXT_PROVIDERS = {}
 
-    def _reg_vcard_plugin(self, what, cfg_sect, plugin_classes, cls, dct):
+    def _reg_vcard_plugin(self, what, cfg_sect, plugin_classes):
         for plugin_class in plugin_classes:
             if not plugin_class.SHORT_NAME or not plugin_class.FORMAT_NAME:
                 raise PluginError("Please set SHORT_NAME "
```
BjarniRunar: We should also check for the HOOKS variable here

mailpile/search.py
```diff
@@ -605,6 +605,9 @@ def _real_scan_one(self, session,
             last_date = long(msg_info[self.MSG_DATE], 36)
             added += 1
 
+        PluginManager.trigger(
```
BjarniRunar: I don't think this is the right place for it; in this place we may be "lazy" which means we're adding metadata about the message to our index but not actually processing the message body. You want the message body to be available when this takes place; that basically means you want to look for where the keyword extractors are invoked. There is already a hook there for plugins to look at incoming mail, so I'm not sure we need a new one for this.

mailpile/plugins/vcard_face.py
```diff
@@ -0,0 +1,59 @@
...
+    CONFIG_RULES = {
+        'active': [_('Enable this importer'), bool, True],
+    }
+    VCARD_TS = 'x-face-ts'
+    VCARD_IMG = ''
+
+    def __init__(self):
+        pass
+
+    def __call__(self, **kwargs):
```
BjarniRunar: This is pretty vague. Why can this not have a proper name and named arguments? I don't think the complete lack of structure is a good code smell.

mailpile/plugins/__init__.py
```diff
@@ -782,6 +776,15 @@ def get_ui_elements(self, ui_type, context):
         return [elem for elem in self.UI_ELEMENTS[ui_type]
                 if context in elem['context']]
 
+    ##[ Pluggable Hooks ]####################################################
+
+    INCOMING_EMAIL = {}
```
BjarniRunar: I don't like how completely unstructured this is, explicit is better than implicit if we want to avoid weird heisenbugs.

Also note the comments about where this is invoked; I don't think this is integrated at the right place at the moment, and I think the functionality you want already exists (but may have a misleading name).

ryneeverett:
I think I've mostly addressed your concerns but I think I also broke mail scanning somewhere along the way. I'll have to come back to this.

ryneeverett:
Fixed and ready for another round of review.

PR #1683, opened 2016-11-22
Launch browser at start-up option

aaronthj16:
This is so that a browser won't start up instantly for command line users and having to close the browser before using the command line

### DIFF
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -80,7 +80,8 @@ def Interact(session):
 
     try:
         prompt = session.ui.term.color('mailpile> ',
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Thanks for doing this! However, making the startup interactive like that is a non-starter, we cannot always assume the user is present and watching the console.

Preventing the browser from launching can already be done in two ways:

1. Unset the DISPLAY variable (on Unix) before launching Mailpile, or
2. set prefs.open_in_browser = false

This could be better documented obviously... :-)

aaronthj16:
Thanks for that :)

PR #1684, opened 2016-11-22
Improve local encryption

BjarniRunar:
This PR updates the "mailpile encrypted storage" format to v2.

Main changes:

1. Use the pycrypto library if available
2. If we have pycrypto, use AES-128-CTR instead of external openssl enc with AES-256-CBC
3. Instead of using MD5 to validate contents, use SHA256 and proper MAC semantics

This is stage 1 of "get rid of external openssl" - stage 2 will be a migration script that finds and reencrypts data stored in the old format. Stage 3 will be using the included crypto.records to move our index out of RAM and properly encrypt the search engine, but that probably is a post-1.0 feature.

The benefit of AES-128-CTR is that it is malleable, a single bitflip will not corrupt the entire file, which means the our encryption no longer hurts reliability (aside from complexity and bugs, but such is life).

The hope that pycrypto would be faster than shelling out to openssl has unfortunately not been realized, so many cycles are spent in Python on base64 en/decoding, buffering and searching for boundaries that this actually performs a bit worse on my machine. Future work may be able to improve that, we'll see...


JocelynDelalande:
Will review (at least reading & testing, reviewing is not the propper word).

I'm thinking about maintaining the doc in sync with current master.

The recommended install method will include install of pycrypto, right ? Ok if I update the install docs to mention that new dependency ?

As-is, does this PR is supposed to break existing data or transition smoothly ?

Ayrx:
Is there a reason why you opted for PyCrypto over pyca/cryptography? PyCrypto is now completely unmaintained (last commit is in 2014) while pyca/cryptography is in active development. AFAICT all the functionality you require is supported by pyca/cryptography.

Disclaimer: contributor to pyca/cryptography.

JocelynDelalande:
@Ayrx Interesting remark indeed.

1. [last commit on pycrypto](https://github.com/dlitz/pycrypto/commits/master) is in June of 2014

2. On the other side, pyca-cryptography does not seem to be packaged in Linux distros (as far as now, you can install mailpile without a « pip install », at least on debian / ubuntu).

So, no ideal choice, but 1. is IMHO really concerning :-/

Ayrx:
@JocelynDelalande pyca/cryptography is packaged both in Debian and Fedora.

JocelynDelalande:
@Ayrx How fast :)

You are right, I didn't found it because of naming (looking for some package named "pyca"-something ), but you are right, there is [python-cryptography](https://packages.debian.org/search?keywords=python-cryptography)

(@Aryx shouldn't you mention packages on the [installation page](https://cryptography.io/en/latest/installation/) of pyca-cryptography ?)

BjarniRunar:
@Ayrx: Things not being updated constantly doesn't mean they're broken, it can also just mean they're stable. But there's no particular reason to prefer pycrypto over cryptography, it was simply the first crypto lib that I tried, appeared to be widely packaged and available, and worked. AES is AES is AES... no? :-)

@JocelynDelalande Regarding existing data, this will transition smoothly. HOWEVER, you can't go back since the old code doesn't understand the new crypto. And if we want to support cryptography, then I need to make sure that the cryptography implementation of AES-128-CTR and the pycrypto implementation of the same are compatible, or switching will break things again.

JocelynDelalande:
> @Ayrx: Things not being updated constantly doesn't mean they're broken, it can also just mean they're stable. But there's no particular reason to prefer pycrypto over cryptography, it was simply the first crypto lib that I tried, appeared to be widely packaged and available, and worked. AES is AES is AES... no? :-)

Well… most of the time it means *dead project* ; that aside, still it's an health indicator.

You are maybe right on AES (except: implementation errors/bugs/flaws) ; but state of the art of crypto evolves, pycrypto does not. It's a bet on future, with the indicators we have  pycrypto getting unmaintained (if not already…) is more likely than the same with pyca-cryptography.

My take on this: keeping deprecated dependencies could sometimes be an understandable choice (due to the cost of switching), but adding a new dependency to an already half-dead lib seems like a bad plan to me.

BjarniRunar:
@JocelynDelalande Where did you get "deprecated" from? Python libs that basically wrap a low level C library don't need to evolve constantly, their dependencies take care of that. A stable non-evolving wrapper around openssl (or libcrypto, or wherever the code actually lives) is exactly what we want.

But sure, cryptography is cool too. They promise to keep their APIs stable (https://cryptography.io/en/latest/api-stability/), so they're not obviously worse than pycrypto. ;-)

I'll see if I can get it to work, but a quick glance at their docs implies that their AES-128-CTR can not easily be made compatible with that of pycrypto, so I'm going to have to write a migration tool to recover my mail as well...  so this'll take a few days.

BjarniRunar:
Just a quick counterpoint here... remember that we're sending data To The Future. It has to travel a long way, along a bumpy road that may flip bits here and there. This is why I am not using Cryptography's Fernet construct; they chose AES-CBC, which destroys data if a bit gets flipped. That's exactly the same "let's just focus on security" mistake that Mailpile made in v1 of the storage format, which I am now trying to correct. :-)

Library evolution and development is exactly what we Do Not Want, once the data is encoded we need to be able to continue reading it unchanged for decades. Bugs or no, we need to be able to read it. pycrypto's API is flexible enough to read data generated by the cryptography lib, but not the other way around. Make of that what you will. :-)

If I had found a performant error correction code library for python (reedsolo is too slow), I'd be using that too; hopefully a future version of the container format will include such a thing...

Ayrx:
@BjarniRunar Unmaintained is dangerous if the library in question contains a bunch of unvetted C code with a publicly available code execution vulnerability [available][0] on the issues page that hasn't made it into a released version yet.

[0]: https://github.com/dlitz/pycrypto/issues/176

BjarniRunar:
@Ayrx, rather than stating the obvious on one hand, and making insinuations that I'll have to go check and verify myself on the other - perhaps just say "unfortunately pycrypto contains a bunch of C code which has known bugs and nobody is fixing them". Hmm? If you know the answer to the questions you pose, please respect our time and be direct.

I'm working on this.

Ayrx:
@BjarniRunar That was in direct response to your statement of

> Things not being updated constantly doesn't mean they're broken, it can also just mean they're stable.

I'm not sure why you are getting snippy about this.

BjarniRunar:
@Ayrx Because you chose to be condescending rather than direct. But thanks for the feedback, it's useful even if the presentation annoys me.

Ayrx:
I apologise if you take it that way. I had no intention of being condescending.

BjarniRunar:
Did a bit of reverse engineering and figured out how to format the nonces so cryptography and pycrypto would do the exact same thing.

Cryptography is now the preferred default, pycrypto works too, both generate the same data.

Thanks for keeping me on my toes guys, let's hope that was the most serious issue... :-P

flanfly:
Hey @BjarniRunar,

I really appreciate your work on Mailpile and your commitment to users' privacy, but this encryption scheme is in not way secure.

# Vulnerabilities
## Encryption
The PR uses AES in Counter mode to encrypt mails. Before encryption a MD5 or SHA-256 checksum is appended, depending on whenever the classes from `streamer.py` or `records.py` are used.

Using Counter mode (CTR) is an exceptional bad idea, especially when not using any MACs. The malleability of CTR allows an attacker to modify encrypted data in case he has knowledge of the plaintext. The dictionaries in `records.py` completely lack any message authentication and thus are vulnerable. The cryptographic checksums of the plaintext do not prevent this. The comments mention that this is intentional to limit destruction in case of bit flips. This is misguided in two ways. First, it's the responsibility of the file system and transport protocol to prevent these and second, cryptographic checksums are not designed for error detecting or correction.

## Authenticy
`EncryptingStreamer` authenticates the plaintext using a MAC that is appended to the ciphertext. For this it concatenates key and plaintext and computes its SHA-256 checksum. This type of MAC is vulnerable to Length Extenstion attacks [1].

## Key Derivation
Mailpile reuses keys for different security goals. The MAC key and the encryption key are the same this means that knowledge of one breaks the whole system. The encryption/authentication keys are saved in a file that is encrypted with the users' passphrase using GnuPG symmetric encryption. OpenPGP does not guarantee that an iterated KDF is used to derive the encryption key from the passphrase. Combined with the fact that users tend to use low-entropy password this makes exhaustive search possible. All possible KDFs used by GnuPG are space-constant [4] which makes them vulnerable to graphics cards. (this is unrelated to the PR at hand).

## General Nitpicking
The means of encryption and authentication used by Mailpile differ depending on the classes used (`records.py` vs `streamer.py`) and so do the primitives used (MD5 vs. SHA-256 & CBC vs. CTR). This makes security evaluations difficult. Also, the mail encryption implemented in Mailpile essentially reinvents OpenPGP.

Again, I really like your project and respect the huge amount of work that goes into Mailpile. I just want to keep you from implementing bad crypto.

# What to do
Please use a proper authenticated encryption mode like AES-GCM. It's fast, ubiquitous and has a proven track record. In case you want to guard the ciphertext against bit-flips add CRC or an ECC to the encrypted stream.
For key derivation please use `scrypt`. It's neither time- nor space-constant and makes exhaustive search extremely expensive. The derived key can be used with GnuPG to decrypt the keyfile. To derive keys for multiple shards in a `EncryptedDictionary` use a proper KDF like HKDF [2] or a CPRNG like those defined in NIST SP800-90a rev1 [3] (except Dual_EC DRBG :wink:).

Please consider using GnuPG to encrypt mails on disk. OpenPGP solves many edge cases and is standardized so others may be able to interoperate with Mailpile.

1: https://en.wikipedia.org/wiki/Length_extension_attack
2: https://tools.ietf.org/html/rfc5869
3: http://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf
4: https://tools.ietf.org/search/rfc4880#section-3.7

BjarniRunar:
Hi @flanfly, thank you VERY MUCH for your review! Allow me to respond...

Note that the code in records.py is still under development, in fact the top of the file specifically says that it needs to be revisited after switching to CTR. It is not currently in use anywhere. So your comments there are well taken and I will make sure to take them into account before I start using the record-based storage. I intend to copy your comments verbatim and put in the top of the file as a reminder for when I revisit that.

### AES-CTR

I do however strongly reject the assertion that dealing with corrupt data is the filesystem's responsibility. I fully expect these files to both cross the network and live on shitty storage such as external USB sticks. If I have to choose between security and reliability, I will choose reliability.

Being able to recover from corruption is MORE IMPORTANT to MORE USERS than the risk that an attacker will hack their mailpile storage and carefully, maliciously flip a bit in an old e-mail.

The risk of bit-flips is real, the risk of an attacker flipping bits in long term storage for malicious ends is something I am happy to document as "out of scope" for our threat model.


### Authenticity

Your analysis of the MAC is incorrect. The MAC in use is constructed like so:

    sha256(key, sha256(key || nonce || content))

The previous incarnation of this code did indeed have the bug of which you speak - that's one of the things I'm fixing in this PR.


### Key Security

Mailpile does indeed reuse the keys for various things.

But I'm going to need some more help understanding how adding more keys will improve the security of the system, since Mailpile will generally need to have all of those keys in RAM at once, and will need to keep all of them for all time (since this is long-term storage). If one falls, all fall. Given those constraints it seems that having one strong master key is going to be easier to reason about and less error prone than creating and keeping track of a whole bunch of different ones.

If you assert that GnuPG's symmetric (password-based) encryption is insufficient for safeguarding the master key, then I'll have to take your word for that and make some changes.


### What to do?

I'll have to fix our master key storage. If a GnuPG dev tells me that GnuPG's internal KDF is not to be relied upon, I am not going to argue that point!

However, you haven't convinced me that AES-CTR is a bad idea. You rejected my concerns and tried to get me worried about a threat that IMO, doesn't exist. So we disagree there.

Adding an ECC to the encrypted stream is however a good suggestion; I had already taken a look at the Python reedsolo library to that end, but it was too slow. Once I find something that I can use, I'll be doing exactly that. Until then, I want exactly the malleability that AES-CTR provides.

Regarding your recommendation that I consider using GnuPG for encryption... I did in fact, for the reasons you cite, it was my first choice when I started this project.  You're quite right when you say that I'm reinventing a lot of OpenPGP stuff. This Mailpile storage format is heavily by OpenPGP. 

Unfortunately, it GnuPG is just too slow. GnuPG *might* work for the large blobs being handled by the streamer, but as I move to small record-based storage it becomes completely unworkable. If you want this to be something developers seriously consider doing, you guys are going to have to turn GnuPG into a library and abandon the fork-gpg-for-everything model...

bersl2:
> I do however strongly reject the assertion that dealing with corrupt data is the filesystem's responsibility.

Rightly so. Most machines still use journaling filesystems which primarily or only protect metadata (ext4, XFS, NTFS, HFS+, etc.) rather than more "modern" filesystems with data integrity features (ZFS, BTRFS). Assuming that the filesystem protects data is factually incorrect at best.

PR #1687, opened 2016-11-24
Change of thread tree symbols to work under Windows and Mac OSX

### LINKED ISSUE #1671: Conversation Tree Start and End not aligned
The round corners forming the end or beginning of a branch in the conversation tree are not aligned with the straight branch sections
screen shot 2016-11-07 at 19 12 00

lorenzschmid:
The thread tree or conversation view uses multiple unicode symbols to represent the tree. Unfortunately they don't work on all OS's. The following symbols are available:

U+250C (corner up)
┌
U+250D (corner up small)
┍
U+2514 (corner down)
└
U+2515 (corner down small)
┕
U+256D (round up)
╭
U+2570 (round down)
╰

As it turns out in Windows 7 the standard monospace font family "Courier New" contains only the following symbols (of the six mentioned above):
U+250C / U+2514
The remaining four symbols get replaced with the corresponding symbol of the "MS Mincho" font family (which uses a different width). Mac OSX seems to have a similar issue. Therefore I replaced the round corners with their angled counterpart.

This fixes #1671.

### DIFF
--- a/mailpile/plugins/search.py
+++ b/mailpile/plugins/search.py
@@ -214,8 +214,8 @@ class SearchResults(dict):
 
     _BAR = u'\u2502'
     _FORK = u'\u251c'
-    _FIRST = u'\u256d'
-    _LAST = u'\u2570'
+    _FIRST = u'\u250c'
+    _LAST = u'\u2514'
     _BLANK = u' '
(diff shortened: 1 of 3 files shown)


BjarniRunar:
Thank you!

Issue #1630, opened 2016-10-01
Mark message as read only manually (not automatically on read, an option)


ferdiga:
Hi
any way to stop Mailpile to mark massages as read automatically
Example 
thunderbird has an option "Automatically mark messages as read"

another way would be to mark it as "todo".

BjarniRunar:
We don't have this option at the moment, if you read a message the "new" tag is removed and the message appears as read in the UI.

This could be added as a feature relatively easily though, I'm marking this issue as "low hanging fruit" in case someone wants to have a go. The way to go about this, in case someone wants to work on this:

1. Add a boolean preference to mailpile.config.defaults (under prefs). For example "prefs.automatically_mark_as_read", defaulting to True.
2. Check for this setting in the _email_view_side_effects() method in mailpile.plugins.search

That should be it!

aaronthj16:
Hi @BjarniRunar I'd like to work on this issue and there is a problem with checking for the setting in the _email_view_side_effects() method because this method is called as a side-effect of the view method so new mail is not automatically marked as read the user will have to view the mail first. So I was wondering if there is a part of the code that assigns the 'read' or 'unread' tag to new mail and this part would check for the boolean preference and automatically assigns a 'read' tag if it is set to true, any thoughts?

BjarniRunar:
Hi @aaronthj16 !

I am afraid I didn't quite understand your question. However, let me try to answer anyway - if things are still unclear please ask again and phrase your question differently.

1. There is already code that assigns the "new" tag - it happens when messages are added to the index. This code does not need to change! If you want to look at this code, the default behaviour is in the `ProcessNew` function in `mailpile/mail_source/__init__.py`.

2. Currently, whenever a message is viewed, the "new" tag is removed as a side-effect (actually, all tags of type "unread" are removed, there might be more than one). And all tags of type "read" are assigned, if any exist (by default there are no such tags).

The purpose of this change, is to change point 2, not point 1. We still want messages to be marked as new as they are received. The change we want, is to make it possible for users to manually disable the automatic side-effect and make marking messages "read" a completely manual process.

I think the use-case is treating e-mail more like a TODO list, so messages are marked "new" (or "todo") until the user manually untags them, no matter how often they are read.

ferdiga:
Hi,
the remark " treating e-mail more like a TODO list " is exactly what is needed in day to day life.
ERP systems like Odoo treat emails like tasks, which can be assigned to different people and different objects.

[...]

PR #1692, opened 2016-11-26
Implementation of feature #1630 (automatically on read option)

aaronthj16:
Hi this is the feature request for the automatically mark as read option, issue #1630


BjarniRunar:
Looks good, thanks for doing this!

PR #1695, opened 2016-11-27
Prevent external redirects. Fixes #1582

### LINKED ISSUE #1582: Open Redirect Vulnerability
1. There is open redirect issue at ( http://localhost:33411/auth/login/?_path=%2F ) .The parameter _path= / leads to the issue.
2.   With reference to the Improve redirector, issue #360 , 
   
   ```
       if '//' not in destination:
                   destination = '%s%s' % (self.server_url(), destination)
   ```

The above 2 lines are not present in the latest version of mail pile I downloaded.When the necessary changes were made, that is the above two lines of code were added to httpd.py and rerun the program again. Mail pile isn't starting up.

BjarniRunar:
Prevent Mailpile's login from becoming an open redirect bouncer for the wider Internet.

### DIFF
--- a/mailpile/auth.py
+++ b/mailpile/auth.py
@@ -121,11 +121,10 @@ class Authenticate(Command):
         path = self.data.get('_path', [None])[0]
 
         # These are here to prevent people from abusing this to redirect to
-        # arbitrary URLs on the Internet. The first assertion blocks normal
-        # http:// and https:// URLs. The second is a catch-all that prevents
-        # any path that looks like it starts with a protocol: foo:/stuff/.
-        assert('://' not in path)
-        assert('/' in path.split(':')[0])
+        # arbitrary URLs on the Internet.
+        assert('://' not in path)          # https://, http://, ftp://, ...
+        assert(path[:2] != '//')           # //www.example.com/
+        assert('/' in path.split(':')[0])  # mailto:user@example.com
 
         if (path and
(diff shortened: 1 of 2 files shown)


BjarniRunar:
I added a few more security fixes to this... there were more ways to play funny games with the login redirector and HTTPD.

BjarniRunar:
Review window timed out, merging!

mailpile/auth.py
```diff
@@ -121,11 +121,10 @@ def _do_redirect(self):
...
         # These are here to prevent people from abusing this to redirect to
-        # arbitrary URLs on the Internet. The first assertion blocks normal
-        # http:// and https:// URLs. The second is a catch-all that prevents
-        # any path that looks like it starts with a protocol: foo:/stuff/.
-        assert('://' not in path)
-        assert('/' in path.split(':')[0])
+        # arbitrary URLs on the Internet.
+        assert('://' not in path)          # https://, http://, ftp://, ...
+        assert(path[:2] != '//')           # //www.example.com/
+        assert('/' in path.split(':')[0])  # mailto:user@example.com
```
JocelynDelalande: IMHO, relying on *urlparse* does the same job (maybe more), is more explicit, robust & better reviewed :

```
from urlparse import urlparse

url = urlparse(path)
assert(not url.scheme and not url.netloc)
```

My too late cents ;-)

BjarniRunar: That's a good idea, thank you. I'll change this and add to a later PR.

PR #1697, opened 2016-11-27
fix Dockerfile (& upgrade ubuntu)

mathroc:


BjarniRunar:
This seems like a good idea. Folks that want an older Ubuntu (if that is anyone's preference) should probably fork this (cc. @DoubleMalt).

Issue #1698, opened 2016-11-27
improve usage with Docker


mathroc:
I'd like to improve the usage of mailpie from Docker. here is what I have in mind:

* [ ] fix Dockerfile so that it's at least able to build & run (#1697)
* [ ] use alpine to make the image much smaller (it's currently 441MB in #1697)
* [ ] fix gpg key creation (at least it failed on my one attempt to create a 4096 one)
* [ ] run mailepie with tor (inside the same image? or env vars link to an external tor node ? both ? etc.)
* [ ] expose web interface on port 80
* [ ] publish on hub.docker.com (auto build of master->latest and tags)
* [ ] add documentation about usage with docker (examples, volumes to mount, upgrade, etc.)

what do you think about this ? anything missing ?

ryneeverett:
I'd add that there probably needs to be a distinction between development and production use cases in the documentation and likely the image/compose file as well.

BjarniRunar:
Pinging @DoubleMalt - do you want to have an opinion on this?

BjarniRunar:
This all sounds worth doing! Thank you!

My only caveat is, I like to request that you hold off on publishing to hub.docker.com until we go into release-candidate mode. We're iterating a bit too fast right now to want to open up to an audience that isn't willing to do development.

Regarding the GPG key generation, that is almost certainly due to a shortage of entropy and Mailpile timing out. I'm not sure how much you'll be able to do about that from the Docker environment, but if you have clever ideas then I'm all ears!

mathroc:
@ryneeverett yes development probably have different needs than production mode. I don't have much knowledge about python dev so I'm mostly going on this for the production version. someone else might have better idea of how to improve developing with docker

@BjarniRunar don't worry I don't plan on publishing on the docker hub myself (I couldn't setup the proper hooks anyway)
ans finally about the GPG key generation. I found out that's it's because the version installed on ubuntu is gpg2 instead of gpg1.4. I'll have to look at how to install this version on the image until #1133 is resolved

DoubleMalt:
I'm happy to join the fray and offer opinions. I'm just in the process of
getting our ARM docker hub ready for primetime ;)

mathroc:
other things to consider:

* [ ] setup on first run so that it can write inside mounted volume
* [ ] setup variable env (language, home, password?, etc.)
* [ ] chown when starting (to fix permissions on mounted volumes)

BjarniRunar:
[...]

PR #1699, opened 2016-11-27
[wip] alpine based Dockerfile

mathroc:
this is a work in progress

working:

* initialize a new mailpile environment
* adding account
* importing emails / contacts
* generating gpg key
* sending messages

image size is currently 101MB (vs 446MB before)

blocking issues:

*  accessing mailboxes (segmentation fault :/)
> Segmentation faultayed welcome message
mailpie_app_1 exited with code 139

I'm not sure how to find out what is causing the segmentation fault. If you have any suggestions, please let me know!
* unlocking pgp key (not sure which password it's asking for and what should happen when it's unlocked)
* signing outgoing messages (because I can't figure out how to unlock the key)
* encrypting outgoing message:
>   File "/Mailpile/mailpile/crypto/gpgi.py", line 442, in writeout
    fd.write(line)
IOError: [Errno 32] Broken pipe

other issues:

* I don't know when this error appeared :
> Failed to load language C
Failed to configure i18n (C). Using fallback.

see #1698

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,35 +1,41 @@
-FROM ubuntu:16.10
+FROM alpine:3.5
 
-# Install dependencies
-RUN apt-get update -y && \
-    apt-get install -y openssl python-imaging python-jinja2 python-lxml libxml2-dev libxslt1-dev python-pgpdump python-cryptography spambayes tor
-
-# Add code
 WORKDIR /Mailpile
-ADD . /Mailpile
 
 # Create users and groups
-RUN groupadd -r mailpile \
-    && mkdir -p /mailpile-data/.gnupg \
-    && useradd -r -d /mailpile-data -g mailpile mailpile
-
-# Add GnuPG placeholder file
-RUN touch /mailpile-data/.gnupg/docker_placeholder
-
-# Fix permissions
-RUN chown -R mailpile:mailpile /Mailpile
-RUN chown -R mailpile:mailpile /mailpile-data
-
-# Run as non-privileged user
-USER mailpile
+RUN addgroup -S mailpile && adduser -S -h /mailpile-data -G mailpile mailpile
 
 # Initialize mailpile
-RUN ./mp setup
-
-# Entrypoint
-CMD ./mp --www=0.0.0.0:33411 --wait
-EXPOSE 33411
+# RUN chown mailpile: /mailpile-data -R && ./mp setup
 
 # Volumes
 VOLUME /mailpile-data/.local/share/Mailpile
 VOLUME /mailpile-data/.gnupg
+
+# Install dependencies
+RUN apk --no-cache add \
+	ca-certificates \
+	openssl \
+	gnupg1 \
+	py-pip \
+	py-imaging \
+	py-jinja2 \
+	py-lxml \
+	py-lockfile \
+	py-pillow \
+	py-pbr \
+	py-cryptography \
+	su-exec
+
+ADD requirements.txt /Mailpile/requirements.txt
+RUN pip install -r requirements.txt
+
+# Entrypoint
+ADD packages/docker/entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+CMD ./mp --www=0.0.0.0:33411 --wait
+EXPOSE 33411
+
+# Add code
+ADD . /Mailpile
+
--- a/packages/docker/entrypoint.sh
+++ b/packages/docker/entrypoint.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+chown mailpile: /mailpile-data/ -R
+
+su-exec mailpile "$@"


JocelynDelalande:
Hi :)

I'm not even using docker but just wondering : is that a good move to switch onto an "exotic" (mind the quotes) system that is not mentioned in [install docs](https://github.com/mailpile/Mailpile/wiki/Getting-started-on-linux) for the default docker image ?

I mean, it adds maintenance (fixing depolyment/env bugs on alpine + ubuntu/debian), and means less testing on debian/ubuntu env (people using docker for dev will *not* test on deb/ubuntu).

As mailpile is intended for large audiences ; I'd prefer keeping the overweight of 300mio and having the default docker env remaining on Ubuntu. Again, that may be a naive opinion, because I'm no docker user…

DoubleMalt:
Docker themselves announced a while ago the they intend to switch their
default images to alpine for the vast improvements of Docker image size.

While it doesn't seem to have hapened (yet), using ALpine for Docker images
is definietly not exotic any more.

BjarniRunar:
Ping @DoubleMalt , @mathroc!

Should this be merged? I don't use docker myself so I don't know what to do with this issue; but I feel bad leaving it unmerged for months like this.

mathroc:
I've rebased and updated the PR, everything seems to be working fine now

@grekko maybe you can test this too and if it's ok @BjarniRunar can merge it ?

grekko:
@mathroc I tested some basic functionality w/ the Dockerfile and could succesfully:
- Start & configure mailpile
- Add an account & generate GPG keys
- Fetch E-Mails
- Send a signed/encrypted E-Mail

Besides these basic functionality tests I don't know that else to consider and would give a thumbs-up for the change 👍

grekko:
@mathroc I had a second look at the packages you install and discovered that the original Dockerfile installs packes like `tor` and `spambayes` which are not explicitly mentioned in the new one.

mathroc:
@grekko thank you for testing, I did not include `tor` because I think it should be run in a separate container. As for `spambayes` I don't remember why I did not install it; what is it used for ? and why is it not a dependency? (in requirements.txt)

anyone have thought about adding or not `tor` and `spambayes` ?

grekko:
> I did not include tor because I think it should be run in a separate container.

Could you elaborate on that? Is that for security reasons? I guess this would require some adjustments on the Mailpile code. The docker orchestration part should be no problem with docker compose I guess.

> anyone have thought about adding or not tor and spambayes ?

I would argue to keep them for now since I assume they are required by Mailpile. Maybe @BjarniRunar has some thoughts on that.

mathroc:
> Could you elaborate on that? Is that for security reasons? I guess this would require some adjustments on the Mailpile code. The docker orchestration part should be no problem with docker compose I guess.

first reason is that I'm used to have one container for one service, that way it's easier to let people use their own custom image of external services. for `tor`, while not that important, adding it in the Mailpile image would mean a larger image and a maintenance burden. but more importantly, users would probably thought that mailpile is "torified", but [the setup is much more complex](https://github.com/mailpile/Mailpile/wiki/Mail(pile)-TORified)

I don't think adjustments on the Mailpile code would be needed as using Mailpile with tor seems to be transparent to mailpile. it connect to an smtp port that is proxified by tor and the webui traffic comes from a tor but, afaik, Mailpile doesn't care about that either

providing an easy way to setup mailpile + tor could be interesting but I don't think it's a priority

grekko:
> first reason is that I'm used to have one container for one service, that way it's easier to let people use their own custom image of external services.

To me the 1-service-per-container argument sounds a little too dogmatic in the case of Mailpile while I agree w/ the benefits of decoupled services generally.

My assumption was that enabling Tor-Support for Mailpile is a first-class feature and should work w/o any further tweaking by the user. If that is not the current reality but a short-term goal I would vote for inlcuding it in the Dockerfile.

If Tor-Support is not really that important it could probably be also excluded. While I am pretty new to the project and cannot tell much about the importance of Tor-Support probably @BjarniRunar can.

BjarniRunar:
Hello!

Without spambayes you have no spam protection, so although the app will run without it, it is pretty important. It is in requirements.txt, but if the operating system provides a package it makes sense to use that IMO.

Tor is optional, but many of the privacy features Mailpile wants to offer depend on it being present, and that is going to increase over time. If it can be included in the default Dockerfile that seems to me like the right thing to do; folks that don't want it for some reason can fork the file, no?

BjarniRunar:
Another point regarding Tor - future versions of Mailpile will be creating and exposing hidden services. I don't know if this is even possible for Mailpile to do if Tor itself is in a different container. And for context, when we build Mac and Windows packages, I do hope we will be able to bundle Tor, for the same reasons.

mathroc:
thank you for the input @BjarniRunar 

alpine does not have its own spambayes package but it should be installed by pip

I can install tor inside the image right now but it will be much more work to start it along with mailpile. the current ubuntu image does not start it either. It'll mean adding a supervisor such as [s6](https://github.com/just-containers/s6-overlay) to start mailpile, tor & dnsmasq + adding whatever is necessary to configure all this through environement variables

I'd rather add tor in a following PR if that's okay (or only install it if it's useful on its own) as I'll have to learn more about tor before being able to add that

BjarniRunar:
Cool! As an aside, this may in fact be a weak argument for using a slightly larger image than Alpine; something Debian based will already have Tor packages that just work. Having a small image is nice, but if it means we have to reinvent too many wheels it may not be worth it.

mathroc:
well with a debian image you will also have to configure a supervisor. as most images, debian images does not start a initd/supervisor by default (it only starts bash) and it's not really possible/easier to use the same as a "full" debian system (systemd) inside an image as the configuration would be very different (no need to mount filesystems /sys/… start networks, tty, etc.)

it's not really starting multiple processes that will slow me down (it will only take a bit of time) it's mainly that I have close to no experience with tor :)

grekko:
@mathroc Ah, now I see what you are referring too. Having tor in the image is only useful if its run when starting the docker container, but currently we just run the mailpile service anyways.

When you mention the supervisor you are referring to the [PID 1 Problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/), are you?

In any case: If the current/previous dockerized Mailpile solution does not support running the Tor-Service it might make sense to not install the package.

Instead there should be an Issue/Chore that targets extending the the dockerized version of Mailpile to properly run Tor in another container and hook it up w/ the mailpile service. What do you think?

mathroc:
> When you mention the supervisor you are referring to the PID 1 Problem, are you?

yep, well, not exactly the same problem but it's closely related. I don't think mailpile itself produces zombie so I didn't add an init (eg: [tini](https://github.com/krallin/tini) but if we want to start multiple processes, we have to introduce a parent to those processed to not let them run headless (and handle restarting or stopping them when appropriate)

> Instead there should be an Issue/Chore that targets extending the the dockerized version of Mailpile to properly run Tor in another container and hook it up w/ the mailpile service. What do you think?

👍 we'll have to look if it's possible to have tor in a separate container. and even if it is, we might want to first run it in the same image for easier deployment (maybe with the option of using an external tor for advanced used)

BjarniRunar:
I'm going to close this in favour of #1804, which AFAICT has incorporated most of these changes and more. Thanks everyone who chimed in, and especially @mathroc for getting the ball rolling.

PR #1700, opened 2016-11-28
Fixed hack in app.py

aaronthj16:
There is no need for the print command between the MoTD and the splash message.

### DIFF
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -147,7 +147,6 @@ class InteractCommand(Command):
         splash = HelpSplash(session, 'help', []).run()
         motd = MessageOfTheDay(session, 'motd', ['--noupdate']).run()
         session.ui.display_result(splash)
-        print  # FIXME: This is a hack!
         session.ui.display_result(motd)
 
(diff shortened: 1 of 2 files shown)


BjarniRunar:
I feel silly being pedantic about such a minor issue... but why would the MOTD command results always start with a newline?  That's no less of a hack, IMO. :-)

BjarniRunar:
If we're fixing hacks, then how about instead of the print, create a `session.ui.display_spacer()` method that just prints a newline (on printable interfaces) and does nothing on others. What do you think?

aaronthj16:
I think that if we were to implement a `session.ui.display_spacer()` method then we probably should just leave the print there IMO. I agree that this is probably a very minor issue but I thought that 'fixing' the hack was by getting rid of the print so that the code looks more uniform. Adding a method between the MOTD and splash results seems to me the same as leaving a print there.

BjarniRunar:
I agree, it's pretty much the same thing since at that point in the code the console is the only possible communication channel. That may change someday, but until then...

PR #1701, opened 2016-11-28
1st draft at sensible behaviour when out of disk space. Fixes #1696

### LINKED ISSUE #1696: Make sure Mailpile stops safely when out of disk space
If Mailpile runs out of disk space, it should notify the user and prevent any further actions (tagging, mail downloading, etc.) until more space is available. This applies both to the back-end processes and the user interface.

This is critical so users don't experience data loss when they run out of disk space.

BjarniRunar:
Changes:

   * Add a sys.minfree_mb variable for configuring our space requirements
   * Under sys.minfree_mb, stop downloading and indexing mail
   * Under sys.minfree_mb/2, make Mailpile completely read-only
   * Add a HealthCheck command, add notifications in the UI

The HealthCheck command is mostly used internally, but can be invoked from
the CLI if the user is curious. It communicates and keeps state in the Event
Log.

This method will probably be expanded later as other "system health" issues
are identified which we want to check for; this could include low RAM or
reminders to the user about making/enabling backups.

### DIFF
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -12,7 +12,8 @@ from mailpile.conn_brokers import DisableUnbrokeredConnections
 from mailpile.i18n import gettext as _
 from mailpile.i18n import ngettext as _n
 from mailpile.plugins import PluginManager
-from mailpile.plugins.core import Help, HelpSplash, Load, Rescan, Quit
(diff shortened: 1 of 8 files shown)


BjarniRunar:
Review window timed out, merging!

PR #1704, opened 2016-11-28
HTML5 sandbox tweaks, composer tweaks

BjarniRunar:
See individual commit messages!

These commits are together because the theme is "fix problems with the message HTML view", everything else was a side-effect.


BjarniRunar:
Note: This breaks composing/replying/forwarding on the command-line, I'll be adding to this next time I have time to sit down and hack a bit more.

Issue #1685, opened 2016-11-23
BrowseOrLaunch Error


aaronthj16:
Running the command `browse_or_launch` for me gives this error:

```Traceback (most recent call last):
  File "/home/aaron/Repo/Git/Mailpile/mailpile/commands.py", line 579, in _run_sync
    return self._finishing(command(self, *args, **kwargs))
  File "/home/aaron/Repo/Git/Mailpile/mailpile/commands.py", line 577, in command
    return self.command(*args, **kwargs)
  File "/home/aaron/Repo/Git/Mailpile/mailpile/plugins/core.py", line 396, in command
    sspec = config.http_worker.sspec
AttributeError: 'HttpWorker' object has no attribute 'sspec'

browse_or_launch error: Failed: browse_or_launch 
```
Not sure if this is a bug or not and if anyone else experiences this error.

Issue #1622, opened 2016-09-21
Do something sensible when users run Mailpile multiple times at once


BjarniRunar:
The current Mailpile doesn't behave in a sane/safe way if multiple instances of the app run at the same time. Data corruption may easily result.

In the short term, we should prevent the user from running more than one Mailpile.

In the longer term, we should detect that Mailpile is running and "attach" to it somehow, so users can have multiple shell sessions on the same pile.

mironoff111:
How about using the Exclusive Lock?

mironoff111:
And another question (I simply don't want to create another one issue because of this silly question): Is Maipile ready for using in production? What about daemonization (systemd service)?

BjarniRunar:
@mironoff111 ... some sort of lock will be needed, yes. And no, Mailpile is not ready for production. :)

JackDca:
It seems that multiple instances of Mailpile can be run simultaneously without loss of data <u>**if**</u> each uses its own profile subdirectory (i.e. ~/.local/share/Mailpile/profile-name on Gnu Linux). I do this all the time by running test instances using profile <span style="font-family:courier new,courier,monospace">default</span> or <span style="font-family:courier new,courier,monospace">dev</span> and localhost:33411 while my production instance (that handles my personal email) runs in the background (24 hours a day) using profile <span style="font-family:courier new,courier,monospace">jack</span> and <span style="font-family:courier new,courier,monospace">localhost:34411</span>.

Use of the same profile subdirectory by two instances could be prevented by use of a lock file. Package <span style="font-family:courier new,courier,monospace">lockfile</span> is already a dependency (required by spambayes) but <span style="font-family:courier new,courier,monospace">pypi.python.org</span> says it is deprecated and recommends <span style="font-family:courier new,courier,monospace">fasteners</span> instead.

There is still a problem if two instances use the same localhost port. If one instance is running and another instance is started, the second instance will terminate e.g.with

File "/usr/lib/python2.7/socket.py", line 228, in meth  
return getattr(self._sock,name)(*args)  
error: [Errno 98] Address already in use

and some additional messages. There are ways of detecting this if all the profile directories are owned by the same user, but on both Gnu Linux and Windows, it is possible for two users to have Mailpile profile directories with permissions that prevent them from seeing each other. This is a bigger problem (see issue #877) but Mailpile should at least terminate cleanly with an explanation to the user.

[...]

PR #1705, opened 2016-11-28
BrowseOrLaunch error patch

aaronthj16:
This is related to issue #1685. I'm not sure if anyone else has experienced this error but I'm making a PR in case someone is. In any case, the `browse_or_launch` command seems similar to the `interact` command so maybe we should remove `browse_or_launch` and just use `interact` to launch a browser instead.

### DIFF
--- a/mailpile/plugins/core.py
+++ b/mailpile/plugins/core.py
@@ -34,6 +34,7 @@ from mailpile.search import MailIndex
 from mailpile.util import *
 from mailpile.vcard import AddressInfo
 from mailpile.vfs import vfs, FilePath
+from mailpile.conn_brokers import Master as ConnBroker
 
 _plugins = PluginManager(builtin=__file__)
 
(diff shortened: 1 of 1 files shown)


BjarniRunar:
You're misunderstanding the intent of this command. This was a first (rather primitive) attempt at solving #1622.

The idea was that the icon on your desktop would launch Mailpile with --browse_or_launch. If it detected that Mailpile was already running, it would launch a browser and exit. If it detected that Mailpile wasn't running yet, it would continue to an interactive session and a browser.

By removing the exit() call, you've broken this and it will no longer perform the intended function.

You may be right that we should be using the ConnBroker there - but only if this command is run after the monkey-patch which blocks unbrokered use of socket.create_connection. The first change, your fix for #1685 also looks correct to me.

aaronthj16:
Thanks for that :) I'll close this PR and issue #1705 since this is a minor thing and just a misinterpreted error :)

PR #1706, opened 2016-11-29
[wip] fix gpg key generation in docker

mathroc:
go back to ubuntu 14.04 to install gnupg 1.4 instead of gnupg 2

not tested, this is not to be merged yet


ConorIA:
Just running this now. Seems to be working a-okay.

PR #1707, opened 2016-11-29
convert box layout for accounts to table. Fixed #1441

### LINKED ISSUE #1441: boxes too small for content in '/browse'
The boxes are too small for my email address. 
the settings icon hides some of the text, and the IMAP entry is completely hidden

Epse:
So this essentially fixes #1441 by converting the layout with the often too small boxes to a simpler table.
I had to change the `settings` macro defined on that page as well to remove all the right-aligned divs, as these don't play nice with tables.

I hope this is up to standards, took a while to push this due to 🏫

### DIFF
--- a/shared-data/default-theme/html/browse/index.html
+++ b/shared-data/default-theme/html/browse/index.html
@@ -91,13 +91,11 @@
     {%- set _url = '' %}
   {%- endif -%}
   {%- if _url -%}
-    <div class="right">
-      <a class='browse-item-settings link-detail auto-modal auto-modal-reload right'
+    <a class='browse-item-settings link-detail auto-modal auto-modal-reload'
          data-icon="icon-settings"
          href="{{ U(_url) }}" title='{{_("Settings")}}: {{info.display_name}}'>
         <span class='{{_icon}}'></span>
-      </a>
-    </div>
+    </a>
   {%- endif -%}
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Thank you for doing this! Does everything "look the same", or is it worth asking for a screenshot?

I'll review and merge as soon as I have some time to focus.

BjarniRunar:
Looks good! Thank you.

BjarniRunar:
Oh, hm, the code looked good but I hadn't realized that you completely did away with the box concept and replaced it with plain table rows that were 100% wide.

I think this looks quite bad - it's very poor use of space on wide monitors, the white-space is all off, and the columns don't line up... this is definitely out of character with the rest of the app now.

Are you interested in working on this some more to make it look a bit nicer? If not, I think that reverting to the boxes, suppressing the overflow and making full path-names available in tooltips might be a better solution.

Epse:
oh, columns did line up in my testing. weird.

Yeah I'll have a look at what we might do. And I agree on the white-space
thing. That table behaves very oddly, without the width setting it is just
300 pixels wide and all the content disappears.
>
>

Epse:
something did go terribly wrong with that padding on the page title..... I am currently looking at the accounts list for comparison

Issue #1693, opened 2016-11-26
Revisit TOFU strategy, taking user feedback into account


BjarniRunar:
For the past year or so, I have been "living the dream", signing all my outgoing mail and opportunistically sending keys to new interlocutors.

This has given a few insights, which should be folded back into our [Security Roadmap](https://github.com/mailpile/Mailpile/wiki/Security-roadmap) and the code itself:

1. Inline signatures, although ugly, do not appear to confuse any recipients, ever. Their main drawback is that cruft accumulates as quoted content gets requoted again and again.

2. Attachments, whether signature.asc or attached keys, are VERY CONFUSING. I regularly get e-mails back from people telling me they couldn't open the attachments I've sent them, even after I've already explained to them what the files are and that ignoring them is safe.

Lessons, potential solutions:

1. This validates the current default preference for inline signatures over PGP/MIME. Although ugly, it's actually more usable. Some automatic clean-up of quoted content would probably be justified.

2. We need to send these attachments less often and/or make them actually useful to the recipient. The confusion is actually bad enough that I've started manually preventing Mailpile from attaching keys; and if a true believer like myself does this, we should expect less motivated users to simply disable the encryption features or abandon Mailpile entirely. This needs to be fixed.

One way to make detached signatures and attached keys "useful" to recipients that don't have PGP, is to modify their contents so they appear as normal text or HTML when opened by the user, and contain some description of what e-mail encryption is and why. IF this sort of thing can be done without breaking compatibility with PGP-enabled mailers, this would be big win and could also advertise the project in a positive way. It will cost some bytes but reduce friction.

BjarniRunar:
Relates to #912.

JackDca:
RFC2440, section 6.2, allows "Armor Headers" after the usual Armor Header Line.
The Armor Header Line is the  "-----BEGIN PGP PUBLIC KEY BLOCK-----\n" or whatever.  The "Armor Headers" are key-value pairs.  One of the defined key values is Comment. There is one blank line after the Armor Headers and before the base64 encoded data.  So it looks as if this will work (but is it OK to have multiple Comments?)

-----BEGIN PGP SIGNATURE-----
Comment: This attachment contains a digital signature.
Comment: With proper software it can be used to prove that this
Comment: message was sent by Jack Dodds and has not been modified.
Comment: For more information see mailpile.is .

[...]

PR #1713, opened 2016-12-01
Improve usability for recipients of signed email and keys

BjarniRunar:
This PR implements 3 things:

1. Attempt to avoid automatically sending PGP keys to recipients who cannot make use of them
2. Reduce recipient confusion by wrapping keys and signature in friendly HTML
3. Stop adding Content-ID headers to our MIME-parts (for now)
4. Fix a silly buglet in the path-redirection security patches (#1695)

Of these, 3 and 4 are just along for the ride. The justification for the other two is discussed in #1693.

The output from change 2 has been tested with mailpile, mutt, KMail and GpgOL - the signature was handled correctly in all cases and the key was handled correctly in all but GpgOL (the developer is looking into it).


BjarniRunar:
Killing this, I forgot a patch and made a mess of things.

PR #1718, opened 2016-12-05
Improvements to Mailpile's TOFU and PGP key handling

BjarniRunar:
This brings Mailpile a big step closer to implementing our TOFU ideas; the UI now says things about signatures that are based on historic data (extracted from the search engine). This lets us differentiate between brand-new and widely-used PGP keys, and lets us warn the user if someone who always signs their mail suddenly stops doing so or suddenly switches keys.

There are also some bugfixes and minor UI tweaks in there.


nwalfield:
Any reason that you found the TOFU support in GnuPG inadequate?

BjarniRunar:
@nwalfield - In short, I'm not yet willing to take the leap and base any core Mailpile functionality on GnuPG features that are only available in 2.1. I am also still very much on the fence about when to automatically import keys onto the user's keychain, which unless I've misunderstood is a prerequisite for the GnuPG TOFU. Please correct me if I'm wrong about that though!

Issue #1711, opened 2016-11-30
Use a proper key derivation function to protect the master key


BjarniRunar:
According to feedback from Kai of the GnuPG project (see discussion on #1684), the way we currently protect the internal mailpile master key is insecure.

Instead of sending the user's Mailpile password directly to GnuPG, we need to run it through a strong key derivation function first.

flanfly:
For the record: The key derivation in GnuPG isn't broken, we just have better means deriving keys from password now then 2007 when the OpenPGP standard was written.

BjarniRunar:
Progress: I've got working code in my local tree that uses PDKDF2 or Scrypt (preferred), depending on the capabilities of the local crypto libraries.

PR #1722, opened 2016-12-08
Use proper KDF for protecting the master key + trivialities

BjarniRunar:
See commit messages - and issue #1711. :-)


BjarniRunar:
Review window timed out. It seems to work... ;-)

PR #1724, opened 2016-12-10
Improvements to setting and changing passwords

BjarniRunar:
This addresses the most pressing password-set/change deficiencies of the app:

1. Makes it possible to change the password after setup
2. Provides basic feedback in the UI about password strength (length only atm)
3. Provide strongly random password suggestions
4. Get rid of any mention of "passphrase" in the UI, just use the word "password" everywhere.
5. Ensure changing the master password logs out all sessions authed using old credentials
6. Prevent the key lock/unlock tool from bricking keys generated by Mailpile itself
7. Make sure that old passwords aren't kept forever on disk after changing (they are kept for a little while, so users can revert the change if necessary)

### DIFF
--- a/mailpile/auth.py
+++ b/mailpile/auth.py
@@ -174,7 +174,7 @@ class Authenticate(Command):
 
             except (AssertionError, IOError):
                 session.ui.debug('Bad passphrase for %s' % session_id)
-                return self._error(_('Invalid passphrase, please try again'))
+                return self._error(_('Invalid password, please try again'))
 
         if user in config.logins or user == 'DEFAULT':
             # FIXME: Salt and hash the password, check if it matches
--- a/mailpile/auth.py
+++ b/mailpile/auth.py
@@ -77,6 +77,11 @@ SESSION_CACHE = UserSessionCache()
 LOGIN_FAILURES = []
 
 
+def LogoutAll():
+    for k in list(SESSION_CACHE.keys()):
+        del SESSION_CACHE[k]
+
+
 class Authenticate(Command):
     """Authenticate a user (log in)"""
     SYNOPSIS = (None, 'login', 'auth/login', None)
--- a/mailpile/auth.py
+++ b/mailpile/auth.py
@@ -299,8 +299,13 @@ class SetPassphrase(Command):
             return self._success(_('Enter your password'), result)
 
         assert(keyid is not None and fingerprint is not None)
+        if fingerprint in config.secrets:
+            if config.secrets[fingerprint].policy == 'protect':
+                return self._error(_('Protected password'), result)
+
         def happy(msg):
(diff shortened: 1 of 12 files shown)


JocelynDelalande:
> Get rid of any mention of "passphrase" in the UI, just use the word "password" everywhere.

A good move, maybe :), ack-grep and my reviewer spoted some occurences of "passphrase" I'm wondering about (did you forgot them ?) : 

```
mailpile/plugins/crypto_gnupg.py
475:           _('This key\'s passphrase will be used to log in to Mailpile')]

shared-data/default-theme/html/partials/head.html
66:      message: '{{_("Back up your encryption key & passphrase")}}',

shared-data/default-theme/html/settings/backup-keys.html
46:  <p>{{_("In the most tragic event of losing the hardware or data that your Mailpile runs on, you may need this master key (and the passphrase) to restore Mailpile from your backups. You do have back ups, right? Hint, hint, wink... ;-)")}}</p>
88:  <p>{{_("It is very important to make a backup of your Mailpile master key and the passphrase which unlocks your encryption keys.")}}</p>
92:    <li>{{_("Enter your PGP passphrase to confirm your identity")}}</li>
101:    <p>Your passphrase: <input type="password" name="pass" value=""></p>
113:      <li>{{_("The PGP passphrase is used to unlock your public encryption key and verify your identity.")}}</li>

shared-data/default-theme/html/settings/set/password/index.html
97:      <div id="setup-passphrase-existing-confirm">
98:        <span id="validation-passphrase_confirm">
100:          <input id="input-setup-passphrase_confirm"
103:                 placeholder="top secret super duper passphrase" tabindex="3">

shared-data/locale/mailpile.pot
41:msgid "Invalid passphrase, please try again"
1469:msgid "Sent passphrase"
2495:msgid "This key's passphrase will be used to log in to Mailpile"
3691:msgid "Oops, wrong passphrase. Try again?"
4345:msgid "Make sure you print or save your keys & passphrase somewhere secure"
6216:"Mailpile runs on, you may need this master key (and the passphrase) to "
6232:" passphrase which unlocks your encryption keys."
6242:msgid "Enter your PGP passphrase to confirm your identity"
6267:"The PGP passphrase is used to unlock your public encryption key and "
```

What do you think about those ?

BjarniRunar:
Thanks for the comments. Two of those were legit; I need to rebuild the mailpile.pot file to pick up the changes, for translators and the PGP unlocker still mentioned passphrases. I'll add to the PR.

The other uses are either low-level gnupg-related CLI stuff (where using passphrase is in-line with the GnuPG world) or obsolete things that just haven't been deleted yet. The former is the main problem with this change; our vocabulary becomes incompatible with that of the GPG. It seems like there is a line where we should stay compatible, but I'm not sure where that line is.

PR #1733, opened 2016-12-18
Added grunt as a dev dependency (rather than grunt-cli)

broder:
Without the grunt dependency, trying to run a grunt task results in `Fatal error: Unable to find local grunt.`

grunt-cli is supposed to be installed globally (`npm i -g grunt-cli`) rather than as a project dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "mailpile",
   "version": "0.4.1",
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Seems reasonable. I don't actually use Grunt (I inherited that stuff), so pardon any bitrot!

PR #1682, opened 2016-11-20
[WIP] Switch to less-elements-old

JocelynDelalande:
less-elements vanished from github ; and the [official site](http://lesselements.com/) does not even
mention a license. So, there is this *less-elements-old* fork remaining ; it's
not under active development (codebase is tiny) but at least it's online and
FLOSS :-)

Current PR is not sufficient, I am [waiting for a release of Rebar](https://github.com/bnvk/Rebar/pull/24#issuecomment-261797756) including the same patch, which also depends on less-elements old.

MP will need to depend on that new rebar.

PR #1734, opened 2016-12-18
Added less-elements dependency to repository

broder:
https://github.com/dmitryf/elements seems to have been removed so I've added the .less file to the repository.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,5 @@
+{
+  "ignoredDependencies": [
+    "less-elements"
+  ]
+}
--- a/bower.json
+++ b/bower.json
@@ -28,7 +28,6 @@
     "jquery-slugify": "~1.0.3",
     "jquery-timer": "*",
     "jquery.ui": "~1.10.4",
-    "less-elements": "master",
     "listjs": "~1.1.0",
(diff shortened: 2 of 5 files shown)


BjarniRunar:
What do you think @JocelynDelalande , is this better or worse than what your stalled solution? :)

JocelynDelalande:
@broder does it fix the issue with rebar too ?

JocelynDelalande:
@broder FYI, @BjarniRunar is refering to #1682 

> @broder does it fix the issue with rebar too ?

Auto-replying : yes, it does, that's what the igoreDependencies is for.

Ok for me then ; I'm not a big fan, but let's face it, we are stuck and front-end is unbuildable as is, so thanks @broder for that pragmatic fix.

@broder maybe

JocelynDelalande:
Ah, it moved on rebar side ! https://github.com/bnvk/Rebar/pull/24#issuecomment-268667402 So i'd suggest keeping the simpler path taken via #1682 for the moment

broder:
:+1:

JocelynDelalande:
@broder thanks for your contribution anyway, and sorry for your time wasted.

Issue #1735, opened 2016-12-19
Exception in http_proxy module


JoEiner:
When I try to display images in an email (most recent git master checked out), the http_proxy fails.
`Traceback (most recent call last):
  File "/Users/johannes/Desktop/Mailpile/mailpile/commands.py", line 581, in _run_sync
    return self._finishing(command(self, *args, **kwargs))
  File "/Users/johannes/Desktop/Mailpile/mailpile/commands.py", line 579, in command
    return self.command(*args, **kwargs)
  File "/Users/johannes/Desktop/Mailpile/mailpile/plugins/html_magic.py", line 163, in command
    len(data))])
  File "/Users/johannes/Desktop/Mailpile/mailpile/httpd.py", line 176, in send_standard_headers
    self.send_header(header[0], header[1])
  File "/Users/johannes/Desktop/Mailpile/mailpile/httpd.py", line 92, in send_header
    self.assert_no_newline(value)
  File "/Users/johannes/Desktop/Mailpile/mailpile/httpd.py", line 86, in assert_no_newline
    assert(re.search(self._NEWLINE_RE, data or '') is None)
  File "/usr/local/Cellar/python/2.7.12_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/re.py", line 146, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer`

The problem is that `request.send_standard_headers(mimetype=contenttype, header_list=[('Content-Length', len(data))])` provides an integer value for content length, which is not accepted by 

Replacing line 86 (inside assert_no_newline method) with `assert(re.search(self._NEWLINE_RE, str(data) or '') is None)` or changing the input of the `send_standard_headers` call resolves the issue.

BjarniRunar:
This should have been fixed by #1736. Thanks for reporting and working on the patch!

PR #1736, opened 2016-12-19
bugfix: re.search only works on string type data, cast any other type…

JoEiner:
I'm not really wise enough to figure out where to cast best... #1735


BjarniRunar:
Thank you!

Just a very minor nit; if data is None, then str(data) == 'None' - which is harmless and will do the right thing in this particular case, but in other places you would have risked breaking the (data or '') logic which explicitly is trying to convert None and False to the empty string. A better patch would have been str(data or '').

Since we're just looking for newlines, either works in this case though, so you "got away with it". Merging! :-)

PR #1738, opened 2016-12-22
Fix `make debian-dev` for Debian

JocelynDelalande:
Detailed in commits messages.

The `make debian-dev` is broken on Jessie without that.


BjarniRunar:
Thank you! I do suspect this breaks some older installations, but I'm not sure I'm too worried. If I encounter problems I'll just fix them...

PR #1746, opened 2016-12-29
Update Mailpile to work with GnuPG 2.1

BjarniRunar:
See individual commits for details...

### DIFF
--- a/scripts/add-gpgme-and-gnupg21-to-venv
+++ b/scripts/add-gpgme-and-gnupg21-to-venv
@@ -1,8 +1,11 @@
 #!/bin/bash
+#
+# See also: https://www.gnupg.org/download/integrity_check.html
+#
 
 GNUPG_DEPS="\
     https://gnupg.org/ftp/gcrypt/npth/npth-1.3.tar.bz2=1b21507cfa3f58bdd19ef2f6800ab4cb67729972 \
-    https://gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.25.tar.bz2=8ab7494e40f80f4138edc9516981bf4afe7d9dbf \
+    https://gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.26.tar.bz2=9a926e7ee6309e539313443555535d49a2a5c9f1 \
     https://gnupg.org/ftp/gcrypt/libksba/libksba-1.3.5.tar.bz2=a98385734a0c3f5b713198e8d6e6e4aeb0b76fde \
     https://gnupg.org/ftp/gcrypt/libassuan/libassuan-2.4.3.tar.bz2=27391cf4a820b5350ea789c30661830c9a271518 \
     https://gnupg.org/ftp/gcrypt/libgcrypt/libgcrypt-1.7.5.tar.bz2=fa485d854748fc06ea041b3057b2de2f12fbc17f \
--- a/scripts/add-gpgme-and-gnupg21-to-venv
+++ b/scripts/add-gpgme-and-gnupg21-to-venv
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+GNUPG_DEPS="\
+    https://gnupg.org/ftp/gcrypt/npth/npth-1.3.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/libksba/libksba-1.3.5.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/libgpg-error/libgpg-error-1.25.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/libassuan/libassuan-2.4.3.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/libgcrypt/libgcrypt-1.7.5.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/gnupg/gnupg-2.1.17.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/gpgme/gpgme-1.8.0.tar.bz2"
+
+[ "$VIRTUAL_ENV" == "" ] && {
+    echo "Please activate your virtualenv first!"
+    exit 1
+}
+echo
+echo "About to download, build and install GPGME 1.8.0 and GnuPG 2.1.17"
+echo "into: $VIRTUAL_ENV"
+echo
+echo "Press ENTER to continue."""
+read
+
+set -x
+set -e
+
+mkdir -p "$VIRTUAL_ENV/gnupg-build"
+cd "$VIRTUAL_ENV/gnupg-build"
+
+for package in $GNUPG_DEPS; do
+    tarball=$(basename $package)
+    srcdir=$(basename $package |sed -e s/.tar.bz2//)
+
+    if [ ! -e "$tarball" ]; then
+        rm -f dl.tmp
+        wget "$package" -O dl.tmp
+        mv -f dl.tmp "$tarball"
+    fi
+
+    [ -e "$srcdir" ] || tar xfj "$tarball"
+    pushd "$srcdir"
+    ./configure --prefix="$VIRTUAL_ENV"
+    make
+    make install
+    popd
+done
+
+set +e
+set +x
+
+echo
+echo "Whew! If you would like to clean up, run this:"
+echo
+echo "  rm -rf '$VIRTUAL_ENV/gnupg-build'"
+echo
+echo "You will also need to set the following environment variable:"
+echo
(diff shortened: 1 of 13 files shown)


scripts/add-gpgme-and-gnupg21-to-venv
```diff
@@ -0,0 +1,66 @@
...
+mkdir -p "$VIRTUAL_ENV/gnupg-build"
+cd "$VIRTUAL_ENV/gnupg-build"
+
+for package in $GNUPG_DEPS; do
+    tarball=$(basename $package)
+    srcdir=$(basename $package |sed -e s/.tar.bz2//)
+
+    if [ ! -e "$tarball" ]; then
+        rm -f dl.tmp
+        wget "$package" -O dl.tmp
```
JocelynDelalande: Is this script intended for testing only ?

- If yes I think it should be said loudly in the help text
- If no, at least, sha1sums of downloads might be checked

scripts/add-gpgme-and-gnupg21-to-venv
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+GNUPG_DEPS="\
+    https://gnupg.org/ftp/gcrypt/npth/npth-1.3.tar.bz2 \
+    https://gnupg.org/ftp/gcrypt/libksba/libksba-1.3.5.tar.bz2 \
```
JocelynDelalande: I suspect the ordering not be good, see my script run failing : https://outils.crapouillou.net/zerobin/?a392b7647324166a#FnObJoiGlH1fZes3vcFklqtL97UmkOaAXEPu94axM9I=

BjarniRunar: Right you are.

scripts/add-gpgme-and-gnupg21-to-venv
BjarniRunar: The links are all https links, so I'm not entirely sure what problems we'd be solving?

scripts/setup-test.sh
```diff
@@ -7,13 +7,18 @@ else
     export GNUPGHOME="$MAILPILE_HOME"
 fi
 
-if [ -d '/usr/local/Cellar/gnupg/1.4.16/bin' ]; then
-    export PATH=/usr/local/Cellar/gnupg/1.4.16/bin:$PATH
+if [ "$VIRTUAL_ENV" -a -e "$VIRTUAL_ENV/bin/gpg2" ]; then
+    GPG_BINARY="$VIRTUAL_ENV/bin/gpg2"
+else
+    if [ -d '/usr/local/Cellar/gnupg/1.4.16/bin' ]; then
```
BjarniRunar: That might need cleaning up at some point, but I'd rather leave it as is for this PR. This was used last time I did a Mac OS X build / test cycle, the next person to actually do some Mac work would be the right person to tweak that.

BjarniRunar:
I think I've addressed the build order, and the script now verifies SHA1 sums.

BjarniRunar:
... I also added a big comment to warn users about the fact that GnuPG 2.1 is not backwards compatible with 1.4 and 2.0, so if people use this with their main pre-2.1 keychain, things might break. Beware of this when testing!

JocelynDelalande:
> I think I've addressed the build order, and the script now verifies SHA1 sums.

great, it works :-)

> ... I also added a big comment to warn users about the fact that GnuPG 2.1 is not backwards compatible with 1.4 and 2.0, so if people use this with their main pre-2.1 keychain, things might break. Beware of this when testing!

Good idea.

Ok, I tested this PR, both with GPG 2.1 (using your build script) and GPG 1.4 (the one on my Debian system).

### GPG 2.1

- 2048 bits Key generation  OK
- Mail Encryption OK
- Signature OK
- Mail Decryption OK
- Key lookup (« find encryption  keys ») KO: never finds any result

### GPG 1.4

Every gpg operation listed above times out (master is ok)

```
TimedOut: Timed out: _expecter
Boo! GET_LINE keygen.email not found in 
[GNUPG:] GOT_IT
[GNUPG:] GET_LINE keygen.name

  File "/home/jocelyn/dev/Mailpile/mailpile/crypto/gpgi.py", line 1514, in run
    gpg.chat(self.gpg_args(), self.run_script, self.main_script)
  File "/home/jocelyn/dev/Mailpile/mailpile/crypto/gpgi.py", line 1303, in chat
    return callback(proc, *args, **kwargs)
  File "/home/jocelyn/dev/Mailpile/mailpile/crypto/gpgi.py", line 1500, in run_script
    self.expect_exact(proc, exp, timeout=tmo)
  File "/home/jocelyn/dev/Mailpile/mailpile/crypto/gpgi.py", line 1489, in expect_exact
    if RunTimed(timeout, self._expecter, proc, exp, timebox):
  File "/home/jocelyn/dev/Mailpile/mailpile/util.py", line 977, in RunTimed
    RunTimedThread(func.__name__, work).run_timed(timeout)
  File "/home/jocelyn/dev/Mailpile/mailpile/util.py", line 966, in run_timed
    raise TimedOut('Timed out: %s' % self.name)
TimedOut: Timed out: _expecter
```

BjarniRunar:
Thanks for the thorough test and the report back! I have identified and fixed at least one regression in the 1.4 support, which I hope to push up later today and may explain some of those issues. Not sure it will explain all of them though. If you don't want to wait, it's a one-liner:

Add the following line:

    args.insert(1, "--batch")

... immediately after the line `args.insert(1, "--verbose")` in `mailpile/crypto/gpgi.py`.

JocelynDelalande:
Ok, tested again. **Everything went ok with GPG 1.4** ; key lookup (and import ?) are still failing with 2.1.

Summary :

### GPG 2.1

- 2048 bits Key generation  OK
- Mail Encryption OK
- Signature OK
- Mail Decryption OK
- Key lookup (« find encryption  keys ») KO: never finds any result
- Key import : ??

### GPG 1.4

- 2048 bits Key generation  OK
- Mail Encryption OK
- Signature OK
- Mail Decryption OK
- Key lookup (« find encryption  keys ») OK
- Key import OK

## Side note

I'm not sure I'm actually testing the decryption (I'm sending a test message which never gets to a smtp, and remains in outbox)

BjarniRunar:
Thank you!

I have myself identified flaws with this - in particular the crypto/streamer invokes GPG directly and I haven't patched it to work with 2.1. So there will be another round of fixes before I merge this.

JocelynDelalande:
> I'm not sure I'm actually testing the decryption (I'm sending a test message which never gets to a smtp, and remains in outbox)

How can I actually test GPG decryption with a*./setup-test.sh*  env (and without actual SMTP server in the loop) ?

BjarniRunar:
One way to do so is by manually adding local files to the pile (using the Browse tool), and then importing keys into the temporary keychain created by the test script. It's a bit of work. Or just adding an IMAP and sending mail to myself. :-)

BjarniRunar:
Merging this, it probably works now. Need moar testings.

cesarandreslopez:
Just tried on latest commited version. When trying to generate PGP KEY Seeing

`Traceback (most recent call last):
  File "/Mailpile/mailpile/crypto/gpgi.py", line 1562, in run
    gpg.chat(self.gpg_args(), self.run_script, self.main_script)
  File "/Mailpile/mailpile/crypto/gpgi.py", line 1347, in chat
    return callback(proc, *args, **kwargs)
  File "/Mailpile/mailpile/crypto/gpgi.py", line 1548, in run_script
    self.expect_exact(proc, exp, timeout=tmo)
  File "/Mailpile/mailpile/crypto/gpgi.py", line 1536, in expect_exact
    if RunTimed(timeout, self._expecter, proc, exp, timebox):
  File "/Mailpile/mailpile/util.py", line 1010, in RunTimed
    RunTimedThread(func.__name__, work).run_timed(timeout)
  File "/Mailpile/mailpile/util.py", line 999, in run_timed
    raise TimedOut('Timed out: %s' % self.name)
TimedOut: Timed out: _expecter`

GnuPG 2.1 AND also when rolling back to GnuPG 1.4

Using the latest Dockerfile to build (as Feb 16)

largestprime:
I'm getting a red padlock and "missing key" error on all my encrypted emails when testing mailpile on one of my devices, and this error doesn't come up on any of the other devices, and the only difference I can isolate is that the device on which the missing key error occurs has GPG 2.1.18, while the others are still on 1.4.

Is anyone else getting this? It is possible that it is something else... but I can't imagine what.

BjarniRunar:
Hey @largestprime, can you file that as a separate issue? This is going to get lost here, and I do think it needs further investigation.

PR #1750, opened 2016-12-31
Fix CSS for crypto header of inline composer

JocelynDelalande:
There was a CSS bug, the crypto header was floating around in the composer when
composing in the thread/inline view.

For the story, I had to PR #1682 and #1738 on the way to this tiny one-line PR :-)

Before:

screenshot-localhost 33411 2016-12-31 01-23-25

After:

screenshot-localhost 33411 2016-12-31 01-23-47

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -8227,7 +8227,6 @@ tr.modal-tag-picker-item td.checkbox {
   overflow: hidden;
   word-wrap: normal;
   word-break: normal;
-  white-space: nowrap;
 }
 #pile-results td.subject a.item-subject {
   width: 370px;
--- a/shared-data/default-theme/less/app/pile.less
+++ b/shared-data/default-theme/less/app/pile.less
@@ -115,7 +115,6 @@
     overflow: hidden;
     word-wrap: normal;
     word-break: normal;
-    white-space: nowrap;
   }
 
   #pile-results td.subject a.item-subject {


BjarniRunar:
Thank you!

loidor:
I'm curious about what browser this regards. I'm on Chrome and notices no difference at all in this regard. However, it messes with my inbox and subjects longer than 370px ends up wrapping and using more than one line.

If `white-space: nowrap;` needs to be excluded, would it be possible to only exclude it from `#pile-results td.subject` and still have it in `#pile-results td.subject a.item-subject`? That would make it possible to make subject lines end with ellipsis (...) if they're longer than 370px. 

Another way would to completely remove `width: 370px` from `#pile-results td.subject a.item-subject`, and it would wrap automatically if the subject line is longer than 100% of the div's estate.

JocelynDelalande:
> I'm curious about what browser this regards. I'm on Chrome and notices no difference at all in this regard. However, it messes with my inbox and subjects longer than 370px ends up wrapping and using more than one line.

Oops. 

I'm using Firefox  45.

> If white-space: nowrap; needs to be excluded, would it be possible to only exclude it from #pile-results td.subject and still have it in #pile-results td.subject a.item-subject? That would make it possible to make subject lines end with ellipsis (...) if they're longer than 370px.

@loidor Would you PR this in a way that is ok for you ? I'd happily review & test it.

loidor:
I'd love to! Just installed Firefox 45 and see the same problem, so that's good, I guess, since that gives me a testing platform :)

JocelynDelalande:
@loidor Cool :) Don't forget to mention my on your PR so that I get notified :-)

PR #1754, opened 2017-01-03
CSS fix - broken wrapping of subjects and introducing ellipsis...

loidor:
...to shorten too long subject lines. This should also work with Firefox 45.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -8231,6 +8231,9 @@ tr.modal-tag-picker-item td.checkbox {
 #pile-results td.subject a.item-subject {
   width: 370px;
   display: block;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 /* This is because relative % positioning inside a table doesn't work
    * because the table does not yet know its own size.
--- a/shared-data/default-theme/less/app/pile.less
+++ b/shared-data/default-theme/less/app/pile.less
@@ -120,6 +120,9 @@
   #pile-results td.subject a.item-subject {
     width: 370px;
     display: block;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
 
   /* This is because relative % positioning inside a table doesn't work


BjarniRunar:
Thank you!

loidor:
@JocelynDelalande Hah, I'm still new to using github, but hey, that's how to tag :D

JocelynDelalande:
## Review

Good if it fixes the wrapping issue @loidor :)

On the other side, mailpile shows now less information than before (before my patch+your patchs I mean), doing the ellipsis at 370px is loosing space / information IMHO (I want to read long subjects if my screen has space for it).

Example:

screenshot-localhost 33433 2017-01-06 15-27-16

@loidor I don't have time to dig into CSS to give a decent solution to that issue, and you ? (I'd be happy to review and discuss it though).

## Side note

@BjarniRunar I know having a quick feedback is good as a contributor, but I think we should take the time to review/test/discuss PRs.

I don't know exactly what is the best merging/review strategy, but I have the feeling that both #1750 and this #1754 would have benefited for a longer time window for discussion. It's normal that we make mistakes and don't think at everything as individuals, that's why we back each other by reviewing. On the other hand but it's damageable to Mailpile that each time we fix something we break something else :-)

BjarniRunar:
@JocelynDelalande I see where your'e coming from but, I think I disagree.

For cosmetic-only changes like these, I think merging to maximize visibility is the quickest way to flush out the bugs. Waiting for a reviewer to come along, pull the branch and test, and notice problems (which will only happen sometimes anyway, depending on browser, screen size, etc) is unlikely to work but will definitely slow us down.

It's just a gut feeling, but I suspect *most* UI regressions won't be noticed until the changes are merged and get a wider audience. So, I think the sooner the better... but if people put comments in their PR asking me not to merge because they wnat feedback and discussion, I will of course respect that.

loidor:
I kept the 370px because I thought someone had thought something with it sometime. The nicest way would be to switch it out for 90% or something. Then it would stretch all across and still end in ellipsis.

loidor:
The 90% way is on https://github.com/loidor/Mailpile.git, but yeah, the reason I didn't do it in the first place is because I believed the 370px was set for a reason, so someone perhaps have some idea about that before I PR it?

JocelynDelalande:
@BjarniRunar @bnvk any thought on that ?

loidor:
Ah well, I'll look through it once more tonight and make another PR and guess that the 370px thingy isn't the end of the world.

PR #1758, opened 2017-01-10
CSS - made subject line wider before clipping in inbox 

loidor:
After testing out 90% width for a while, I discovered that it broke on some long subject lines but not others - the overflow never came around and the result was a subject line flowing out of screen.

So, now I've set a width of 700px to the subject line. The #pile-results row in whole is right now at a min-width of 800px, and as long as that's the case this seems like a perfect solution. 

Ping @JocelynDelalande if you'd like to test it out.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /* Mailpile - Default Theme
 *  Version 0.4.1
 *	 Designed and built by @brennannovak and others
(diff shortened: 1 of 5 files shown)


BjarniRunar:
Actually, I am pretty sure some of the 370px stuff was there for exactly that reason, so things would work on smaller screens. If you're expanding the default, could you at least take a look in tablet.less and mobile.less and add back some narrower constraints there?

loidor:
I've added the 370px to tablet.less and mobile.less, so it should be good to go now.

BjarniRunar:
Thanks! It doesn't look like those changes made it into default.css though.

Sorry to keep nitpicking... I do want to merge this! ;-)

loidor:
Don't worry, I've never done anything like this with css before so all nitpicking is appreciated :)

BjarniRunar:
In part I'm concerned that since I didn't see the changes in default.css, then that suggests you may not have tested them and we don't really have any reason to expect them to work... :-)  But also, if they're not in defualt.css then nobody benefits until the CSS is rebuilt.

loidor:
Aah, so the less files are source and the css is the compiled product? I've never worked with less before, so, well, yeah. I simply manually put the changes in both the less files and in the default css file, since it was kind of obvious where it should go. Here, I have no idea what to do. :/ Sorry 'bout that.

So, how should I continue from here?

JocelynDelalande:
> Aah, so the less files are source and the css is the compiled product? 

Yes, and both are commited in the repository (quite uncommon, but that's the straightforward way it work actually).

> I've never worked with less before, so, well, yeah. I simply manually put the changes in both the less files and in the default css file, since it was kind of obvious where it should go. Here, I have no idea what to do. :/ Sorry 'bout that.
> 
> So, how should I continue from here?

Normal process is making the edits in the less then compile and check the result.

Here is a starting point : https://github.com/mailpile/Mailpile/wiki/Frontend-Dependencies 

Once dependencies are installed using this doc, a `make less` will do the job :)

Don't be frightened at less, it's yet another tool, but I find its learning quite quick and easy.

loidor:
That was a much easier way to do it. I noticed it changed a heap of things back and forward in the css though, and broke a lot of things, so I'll close this, figure out how to do this properly and get back once it's all working properly.

loidor:
So - short question: When I compile less using `make less`, all relative directories get wonky - i.e. what's supposed to be `../` becomes `../../`. Am I doing something wrong? 

If I change webfonts.less and icons.less and replace `../../` with `../`, everything seems to work out perfectly. Is this a known error?

JocelynDelalande:
> So - short question: When I compile less using make less, all relative directories get wonky - i.e. what's supposed to be ../ becomes ../../. Am I doing something wrong?

What is your current working directory when you issue the `make less` ?

loidor:
I'm in mailpile root directory. If I enter any other directory (i.e. default-theme) it responds with "make: Nothing to be done for 'less'."

JocelynDelalande:
@loidor, what OS/Distro are you using ?

loidor:
@JocelynDelalande Debian Stretch on a raspberry pi 3. I could give it go on Mint tonight, and we'll see if that's the problem.

JocelynDelalande:
@loidor : I think you have the wrong less compiler installed (the nodejs one and not the ruby one) ; I think you should :

- Remove the less compiler you have installed if it's the nodejs one
- Install mailpile dev dependencies via `make debian-dev`

There might be  some effort to be done on documentation on that topic.

loidor:
Thanks! I completely uninstalled nodejs (which I had since earlier) and ran `make debian-dev` again, and now I'm back on :)

PR #1765, opened 2017-01-18
CSS - subject line width adaptive to screen size

loidor:
So, now I've learnt how to use less and everything is working fine. Subject line width is set to 700px in default and to 370px in tablet and mobile.

@JocelynDelalande @BjarniRunar


JocelynDelalande:
Superseeds #1758 . @loidor that is totally fine ; however, note for later, you could have kept the same PR just rewriting the commits. Just to tell you that it's a possible thing :)

loidor:
Ah, didn't know that. Ah well, I like learning :)

BjarniRunar:
Thank you!

PR #1766, opened 2017-01-20
CSS - Made body overflow hidden...

loidor:
…and thereby fixes the login page growing out of screen size when logging in. I couldn't find an issue about this and didn't know if I should make a new one or not since it was such an easy fix. Anyhow, I can't see that it breaks anything, but please test it since it affects body and thereby everything.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -6288,6 +6288,10 @@ div.motd a.motd-signed {
   line-height: 14px;
   color: #ffffff;
 }
+/* Body */
+body {
+  overflow:hidden;
+}
 /* Content */
 #content {
   position: fixed;
--- a/shared-data/default-theme/less/app/global.less
+++ b/shared-data/default-theme/less/app/global.less
@@ -1,3 +1,8 @@
+/* Body */
+  body {
+    overflow: hidden;
+  }
+
 /* Content */
 
   #content {


JocelynDelalande:
Thanks for your contrib :-)

Please squash your commits into one.

loidor:
Hah, now I'm there again, not knowing enough of git to do this properly. So, it's easier for me to delete the branch and start over, since I've more than likely screwed up something along the way. That would however mean I can't use this pull request, since the branch would be deleted, I think. 

Would that be better/easier? If I started over?

loidor:
Look at that :) I think I pulled it off!

JocelynDelalande:
(sorry for the delay @loidor )

Ok, on the funtional point of view, that's perfectly fine (I tested locally and I don't see any regression so far). Thanks!

However, on the enlightning road to git, you still have one step left  :-) : using properly commit messages.

You have to imagine that your commit messages will end up "as-is" in the main mailpile git history. Thus, your commit message, « Redid everything » is not telling anything about what you actually wanted to do in that commit.

JocelynDelalande:
(you can amend your commit message locally (git commit --amend), and then push --force it to your "login" branch, it will update the PR automatically.

loidor:
Hah, I didn't think about that, no :) I'll fix it as soon as I can!

PR #1767, opened 2017-01-22
Fixes overflow issue with very long mail addresses in notification box

loidor:
When using very long email addresses, these overflowed the Notifications box. This takes care of that and makes them end with ellipsis instead of overflowing.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -6288,6 +6288,10 @@ div.motd a.motd-signed {
   line-height: 14px;
   color: #ffffff;
 }
+/* Body */
+body {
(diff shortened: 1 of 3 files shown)


BjarniRunar:
Cool! Thank you.

BjarniRunar:
@loidor Just FYI, I am going to have to revert this and wanted to let you know why: this patch limits the notification message to one line only. Many notifications are long and span multiple lines, this patch prevents critical information from being shown. Some other fix will be needed for the overlong e-mail addresses.

loidor:
That's cool, I'll have a second look if I can do some js magic with it instead :)

PR #1775, opened 2017-02-07
Miscellanious fixes

BjarniRunar:
This PR is a mixed bag of random bugfixes and minor tweaks which haven't fitted in any of the larger thematic PRs of late... and I'm too lazy to file each as a separate PR. :-)

Probably best to review each commit separately.

### DIFF
--- a/mailpile/crypto/aes_utils.py
+++ b/mailpile/crypto/aes_utils.py
@@ -99,7 +99,10 @@ def make_dummy_utils():
 try:
     aes_ctr_encryptor, aes_ctr_decryptor = make_cryptography_utils()
 except ImportError:
-    aes_ctr_encryptor, aes_ctr_decryptor = make_pycrypto_utils()
+    try:
+        aes_ctr_encryptor, aes_ctr_decryptor = make_pycrypto_utils()
+    except ImportError:
+        raise ImportError("Please pip install cryptography (or pycrypto)")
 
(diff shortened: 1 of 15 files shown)


mailpile/crypto/aes_utils.py
```diff
@@ -99,7 +99,10 @@ def aes_ctr_decryptor(key):
 try:
     aes_ctr_encryptor, aes_ctr_decryptor = make_cryptography_utils()
 except ImportError:
-    aes_ctr_encryptor, aes_ctr_decryptor = make_pycrypto_utils()
+    try:
+        aes_ctr_encryptor, aes_ctr_decryptor = make_pycrypto_utils()
+    except ImportError:
+        raise ImportError("Please pip install cryptography (or pycrypto)")
```
karlp: could be more verbose, like, "Couldn't find python 'cryptography' module, nor fallback 'pycrypto' module" ?  but yeah, this is already going to be better. :)

BjarniRunar: I like short messages, they're easier to parse since they fit on one line. ;-)  This is only affects the tiny group of users who installed Mailpile >2 months ago and are upgrading after this gets merged, so not worth spending too much time on. All recent/new installs should pick up cryptography from requirements.txt... so not worth polishing too much.

Issue #1778, opened 2017-02-11
Docker image build fails


viossat:
The commit 65884c2c011b759660cb84cd3e604752c89fabc5 breaks the build of the Docker image:
```
Step 10/14 : RUN ./mp setup
 ---> Running in 79eb02d887bc
Traceback (most recent call last):
  File "./mp", line 13, in <module>
    from mailpile.app import Main
  File "/Mailpile/mailpile/app.py", line 9, in <module>
    from mailpile.commands import COMMANDS, Command, Action
  File "/Mailpile/mailpile/commands.py", line 15, in <module>
    from mailpile.eventlog import Event
  File "/Mailpile/mailpile/eventlog.py", line 9, in <module>
    from mailpile.crypto.streamer import EncryptingStreamer, DecryptingStreamer
  File "/Mailpile/mailpile/crypto/streamer.py", line 50, in <module>
    from mailpile.crypto.aes_utils import getrandbits
  File "/Mailpile/mailpile/crypto/aes_utils.py", line 105, in <module>
    raise ImportError("Please pip install cryptography (or pycrypto)")
ImportError: Please pip install cryptography (or pycrypto)
The command '/bin/sh -c ./mp setup' returned a non-zero code: 1
```

Reproduction:
```
docker build -t mailpile .
```

BjarniRunar:
Quoting your message:

    ImportError: Please pip install cryptography (or pycrypto)

... is the Docker image manually installing dependencies instead of using requirements.txt?

BjarniRunar:
Does adding python-cryptography to the apt-get install line near the top of the Dockerfile resolve the issue?

viossat:
requirements.txt is not used, pip is not even installed on the image.
Installing package `python-cryptography` solves the issue, see #1779.

PR #1779, opened 2017-02-11
Add package python-cryptography

viossat:
Fix mailpile/Mailpile#1778


BjarniRunar:
Thanks for doing this!

Issue #156, opened 2013-09-05
Protect message headers in encrypted mail (renamed)


danx0r:
Old subject: Optional encrypted subject line

if the option is checked, the unsecured subject header would be something generic like "subject:ENCRYPTED", and the real subject would be in the text of the email, possibly an extra MIME part.  Mailpile would gracefully use the encrypted subject when possible, suppressing it from the body and showing it in the subject field (and threading on it of course)

Just a thought

BjarniRunar:
Update: This work is now happening as part of the Memory Hole initiative, http://modernpgp.org/memoryhole/ and some code exists in `mailpile/contrib/experiments`

ralfhauser:
To have an overview of encrypted mailing lists see

https://systemausfall.org/wikis/howto/CryptoMailingLists

BjarniRunar:
This is incoming! As a quick reference to the implementation I've ended up with, here is a super-brief discussion of how Mailpile will implement header protections (Memory Hole)...

## E-mail generation

1. When Mailpile generates a PGP/MIME encrypted or signed message, a selection of standard rfc2822 headers (From, To, Cc, Subject, ...) is copied to the header section of the outermost MIME part within the PGP-protected envelope. These copies are considered **protected**.

2. When Mailpile generates a PGP/MIME encrypted message, a subset of the *protected headers* will be **obscured** from the main message header. Obscuring means different things for different headers; for Subject it means replacing the subject with the words "Encrypted Subject", for From/To/Cc it means stripping names and comments (leaving e-mail addresses).

3. The process of obscuring headers can be tuned for more privacy, at the expense of less compatibility with e-mail apps that don't implement Memory Hole. As an example, obscuring the `in-reply-to` header would be good for privacy, but would break threading in many apps. Mailpile will offer one public setting in the UI "minimize metadata" which will choose a slightly more aggressive policy than the default. The default is to only obscure the Subject.

4. When headers are obscured, Mailpile will add a `text/rfc822-headers` part (with the attribute protected-headers=v1) as the first visible part (content-disposition: inline) within the encrypted section. This part will contain a copy of the original (unobscured) headers, so users with mail user agents that don't support Memory Hole can still see the content.

## E-mail parsing

1. When parsing a PGP/MIME e-mail, any non-MIME headers found in the outermost MIME part are copied to the main rfc2822 message header (overriding anything that was there before).

[...]

PR #1783, opened 2017-02-15
Implement message header encryption/signatures

BjarniRunar:
This PR implements the "Memory Hole" spec for encrypting or signing message headers.

Memory Hole signing is enabled by default for all PGP/MIME messages (signed or encrypted) and encrypted messages will also have the Subject line obscured and moved into the encrypted section.


BjarniRunar:
Please see my most recent comment on #156 for details on what this is doing, how and why.

PR #1793, opened 2017-03-01
Change cursor on suggest button

balping:
... on /setup/password


balping:
I'll do some other similar things, closing until then

PR #1798, opened 2017-03-07
Docker: Mount code + config inside the docker container

grekko:
I reallly like to work locally w/ all Mailpile code being executed in the docker container.

The sourcecode and mailpile data I'd like to keep on the docker host instead and mount it into the container, in order to: 
- have file changes immediately be reflected in the running application
- persists config when shutting the docker container down

With my change I am able now to check out the Mailpile-project and start working with a simple:

```
docker-compose up
```

Now I can edit the sourcecode on my machine and see my (front end code) changes immediately reflected inside the running application. For seeing a change in the python code to take effect I still need to restart the docker container.

Another benefit is that I keep my `/mailpile-data` and can restart the docker container w/o having to walk through the setup.

---
 
If there is anybody else also working on Mailpile w/ Docker I'd love see if they are able to work with my changes.

---

I could not find an easy way in docker/docker-compose to configure the permissions for the `/Mailpile` and `/mailpile-data` directories so that the `mailpile` user inside the docker container has proper read-write access. To work around the issue I use `usermod` to set the `mailpile` uid
to 1000.

The issue is documented inside the `Dockerfile`.
- https://github.com/docker/docker/issues/7198
- https://denibertovic.com/posts/handling-permissions-with-docker-volumes/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,34 +2,52 @@ FROM ubuntu:16.10
 
 # Install dependencies
 RUN apt-get update -y && \
-    apt-get install -y openssl python-imaging python-jinja2 python-lxml libxml2-dev libxslt1-dev python-pgpdump python-cryptography spambayes tor
+    apt-get install -y openssl \
+                       python-pip \
+                       python-imaging \
+                       python-jinja2 \
+                       python-lxml \
+                       libxml2-dev \
+                       libxslt1-dev \
+                       python-pgpdump \
+                       python-cryptography \
+                       spambayes \
+                       tor
 
-# Add code
-WORKDIR /Mailpile
+# Add mailpile code (required for initial setup, not for develpoment)
 ADD . /Mailpile
 
-# Create users and groups
-RUN groupadd -r mailpile \
-    && mkdir -p /mailpile-data/.gnupg \
-    && useradd -r -d /mailpile-data -g mailpile mailpile
+# Create data dir
+# This will be overriden by a volume hosted by the docker host (your dev machine)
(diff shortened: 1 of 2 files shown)


BjarniRunar:
We now have two in-flight pull requests for improving the Dockerfile, this one and #1699. I don't really know how to evaluate the quality of either. :-(

However, if I understand it correctly, the use case you describe here is significantly different from that handled by the other. Do we need two Dockerfiles for different modes of operation? Can Dockerfiles take arguments?

grekko:
I imagine @mathroc started working on the alpine-based docker image to decrease the docker image size and maybe to reduce the memory/cpu-usage (I am just guessing here)?

If the alpine version of a dockerized Mailpile achieves that while running stable and without substantial compromises that sounds like a worthy goal.

> However, if I understand it correctly, the use case you describe here is significantly different from that handled by the other. Do we need two Dockerfiles for different modes of operation? Can Dockerfiles take arguments?

I would actually hesitate from creating more than one `Dockerfile`. Assuming someone is using docker for developing Mailpile, a shared `Dockerfile` for dev and the dockerized installation would offer the benefit of dev prod parity. To me a good reason to try share the `Dockerfile` for both use-cases.

I imagine there is a-way™ to build a `Dockerfile` that is on one hand used for development and on the other hand used for releasing a dockerized installation of Mailpile but I have never tried to achieve that.

That being said I'll research wether I can adapt the work of @mathroc in #1699 into my PR and re-use the same `Dockerfile` for dev and prod.

grekko:
Closing in favour of #1804

Issue #1797, opened 2017-03-07
Compose: Persisting user selection (signing/encryping) when adding/removing recipients


grekko:
I am not entirely sure if this is a bug but it did feel odd to me, so I will try to make my point:

**Step 1: Empty recipients list**

When composing a new email I start w/ an empty list of recipients:

![Step 1: Empty recipients list.](https://dl.dropboxusercontent.com/s/re4ai7xstvm4y2i/Screen%20Shot%202017-03-07%20at%2014.30.46.png)

**Step 2: Adding the first recipient**

I add `anna@example.com` and Mailpile enables the option for signing the e-mail I am about to write.

![Step 2: Mailpile enabled the signing option](https://dl.dropboxusercontent.com/s/5cyesjlvnfqi7mc/Screen%20Shot%202017-03-07%20at%2014.31.06%20%281%29.png)

**Step 3: Disabling the signing option**

I decide I do not want to sign the email and disable the option.

![Step 3: I decide I do not want to sign the email and disable the option.](https://dl.dropboxusercontent.com/s/qpa82qtalx2gx8s/Screen%20Shot%202017-03-07%20at%2014.31.24%20%282%29.png)

**Step 4: Adding another recipient re-enables the siging option**

When I now add another recipient `bob@example.com` Mailpile re-enables the signing option again.

![Step 3: The signing option is automatically re-enabled again.](https://dl.dropboxusercontent.com/s/2fcz4392bpkyqi1/2017-03-07%20at%2014.32.png)

---

I would argue that Mailpile should respect my decision to now sign this email and keep the option like that even though when I am adding/removing recipients.

BjarniRunar:
I agree. Mailpile should remember and respect your decision - whenever possible.

There are however combinations which are impossible; if you add a recipient for which you have no key, we can't keep the lock green.

grekko:
Agreed. I changed the UI right now only so much that the Signing-Feature will be persisted.

grekko:
This can be closed since #1799 got merged 👍

PR #1799, opened 2017-03-07
Composing E-Mails: Keep UI and user selection for encrypting/signing in sync

### LINKED ISSUE #1773: (trivial) tooltip on send still shows "signing" when signing disabled
Clicked the "signing" icon to turn it off, and noticed that the tooltip on the Send button still says that it will be signing the mail.  See attached image.

mailpile-201702-signing-not-signing-tooltips

grekko:
My attempt to resolve  #1773 #1797.

Since I changed the python code I am not sure wether I need to adjust some tests or if that change is OK at all. I'd really like some feedback on that part.

I thought about implementing some basic unit tests but could not find any setup so far. If there are no tests yet I might skip them for now.

### DIFF
--- a/mailpile/plugins/crypto_policy.py
+++ b/mailpile/plugins/crypto_policy.py
@@ -217,16 +217,15 @@ class CryptoPolicy(CryptoPolicyBaseAction):
                         break
                 if should_encrypt:
                     policy = 'sign-encrypt'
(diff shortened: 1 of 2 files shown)


BjarniRunar:
Thanks for this! I'll have to review it carefully to make sure you're not persisting impossible states. :-)

PR #1804, opened 2017-03-16
Docker: Mount code + config from docker host

grekko:
This updates #1798 respecting the changes in #1699

It uses two seperate Dockerfiles, one for running the application (`Dockerfile`) and one for development (`Dockerfile.dev` which builds upon the former).

The main goal of this change is that the config and code of the application are not embedded into a docker volume but mounted from the docker host. The `docker-compose.dev.yml` does exactly that.

To start a development environment one can run:

```
docker-compose -f docker-compose.dev.yml up --build
```

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -12,5 +12,5 @@ services:
       - .:/Mailpile
       - .dev-mailpile-data:/mailpile-data:rw
     ports:
-      - 33411:33411
+      - 33412:33411
 
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,6 +1,8 @@
 version: '3.0'
 services:
   mailpile_dev:
+    tty: true
+    stdin_open: true
     container_name: mailpile_dev
     build:
       context: .
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,0 +1,14 @@
+version: '3.0'
+services:
+  mailpile_dev:
+    container_name: mailpile_dev
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
+    image: mailpile_dev
+    volumes:
+      - .:/Mailpile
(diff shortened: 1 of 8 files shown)


BjarniRunar:
This is marked as "wip" - does that mean you don't want me to merge it just yet? I'm otherwise inclined to do just that, and then Docker folks can iteratively improve on it...?

As we discussed the other day, one obvious improvement would be launching Mailpile inside a screen session so the user can access the CLI. But I don't think this PR should be blocked for that.

mathroc:
> As we discussed the other day, one obvious improvement would be launching Mailpile inside a screen session so the user can access the CLI. But I don't think this PR should be blocked for that.

I don't think this is useful, one can always use `docker exec` or `docker-compose exec` to run sh inside a running container

docker-compose.dev.yml
```diff
@@ -0,0 +1,16 @@
+version: '3.0'
+services:
+  mailpile_dev:
```
mathroc: why not having this one in `docker-compose.yml` too ?

grekko: Maybe I missed it. Will have a look at it later this evening.

grekko:
> This is marked as "wip" - does that mean you don't want me to merge it just yet? I'm otherwise inclined to do just that, and then Docker folks can iteratively improve on it...?

I prefixed it `wip` because I am not entirely confident about my approach for the docker-based dev env, but I agree that people can just iterate on my proposal. So, please do not hesitate to merge it.

grekko:
> As we discussed the other day, one obvious improvement would be launching Mailpile inside a screen session so the user can access the CLI. But I don't think this PR should be blocked for that.

I actually tried to get that running but stumbled across mutille problems and discovered that `docker attach' does exactly that. For convenience I wrapped it into the `scripts/docker-dev/shell` script.

grekko:
Rebased against latest master branch.

PR #1807, opened 2017-03-22
Display/Fix current keybindings and provide defaults based on gmail

grekko:
I am heavily using the keyboard for organizing my email and wanted to give the current keybindings a little facelift.

~~The code is based on my already [existing PR](https://github.com/mailpile/Mailpile/pull/1804) for a docker-based dev. In case the merge for that PR will take some more time I would rebase this one on the current master.~~

**This PR**:

1. Updates the keybinding library `Mousetrap` to its latest version 0d792d5
1. Removes some dead code b45d146
1. Refactors some internal functions for accessing the modal element ade67a4
1. Adds a modal for displaying the current keybindings by pressing "?" and configures the gmail defaults 83b2334
1. Tweaks the styling of the modal and refactors the css a little f967c34

After pressing "?" the following modal will be displayed:
keybindings modal

Since this PR might be changing the defaults for some basic controls (e.g. archiving an email, marking an email as read/unread) I am open for discussing wether we should stick with the current settings or we are able to go with gmails defaults.

My reasoning for following the gmail keyboard shortcuts is that in my perception (please correct me here) gmail was the first webmail client with sophisticated keyboard shortcuts and kind of shaped many peoples muscle memory (at least mine 😁 ) for managing emails.

I am grateful for any comment on that.


grekko:
What I am still missing is the information that pressing "?" will display the keybindings.

I thought about something like this:
Did you know? Shortcuts!!

shared-data/default-theme/html/jsapi/templates/modal-display-keybindings.html
```diff
@@ -0,0 +1,25 @@
...
+        <tbody>
+          <% _.each(keybindings, function(binding) { %>
+            <tr>
+              <td><%= binding.title %></td>
+              <td><%= binding.keys %></td>
+            </tr>
+          <% }); %>
+        </tbody>
+      </table>
+      <p>Keybindings are based on <a href="https://support.google.com/mail/answer/6594?hl=en" target="_blank">Gmails keybindings</a>.</p>
```
BjarniRunar: There is a typo here - It should say Gmail's (with the apostrophe).

Also, the message itself, as well as the Keybindings, Title and Binding headings need to be translatable.

Accommodating i18n may mean rephrasing things slightly, because the sentence should be translated as a single unit (translating fragments leads to poor translations due to grammatical differences between languages), while the link markup should not be included in the translated data (so translators don't introduce typoes that break the layout/url). This actually gets awkward enough that it may be best to just skip the attribution entirely.

shared-data/default-theme/html/jsapi/search/ui.js
```diff
@@ -79,6 +75,14 @@ Mailpile.results_list = function() {
 };
 
 
+Mailpile.display_keybindings = function(elem) {
+  Mailpile.API.with_template("modal-display-keybindings", function(modal) {
```
BjarniRunar: Would it make sense to check here if there is already a modal active, and do nothing in that case? We don't want to clobber an ongoing action just because the user hit ? by accident. Since our current UI can only have one modal active at a time, this becomes a problem.

shared-data/default-theme/html/jsapi/templates/modal-display-keybindings.html
grekko: Agreed. Lets keep it simple.

shared-data/default-theme/html/jsapi/search/ui.js
grekko: Good catch. I agree the modal should follow the Highlander-Pattern preventing subsequent calls modals while one is open.

grekko: I'll come up w/ a proposal.

grekko:
@BjarniRunar 

I updated the PR accordingly.

On a side note: This is nothing I would want to address in the scope of this PR but maybe in another refactoring. If we could assume that just every UI string is translatable we might just ditch all those repetetive `{{_("Mark as read")|escapejs}}`-calls and put them inside the template code. What do you think?

grekko:
One more thing regarding coding conventions: I am used to bumpy case naming for variables and methods in JavaScript. So naturally I would name methods not:

```javascript
Mailpile.display_keybindings
```

but

```javascript
Mailpile.displayKeybindings
```
It seems the current coding convention for variables and method names is snake case though, so I wanted to keep it consistent. Do you have any opinions on that one? :)

BjarniRunar:
Not sure I understand what you mean by "put them inside the template code" above - however, you may be missing the fact that the syntax is repetive for a reason: the tools which build the translation catalogues grep the code for _("...") constructs. Forking those tools and changing is a rabbit hole I'd rather not go down. It's true that at the python layer we could change that to include escapejs functionality and reduce the biolerplate somewhat (and avoid mistaken omissions), but that conflicts with all the non-JS use-cases of the exact same function.

Regarding naming: I just want things consistent. Renaming everything to make it look different doesn't seem like good use of anyone's time (and it's hard and annoying to review).

BjarniRunar:
Also, more regarding translation - if you are suggesting moving the strings out of the JS and into templates, that just makes the code harder to understand. Strings should be near the code they are describing so they are more likely to get updated together.

This is in fact the biggest issue I have with the current organization of our JS code - it's broken up somewhat arbitrarily by "what kind of functionality is this" rather than being organized by "what is this code doing". So you almost always have to open up and keep track of multiple files when you want to understand something or make changes. This appeals to people that like certain kinds of structure, but it actively reduces readability (and maintainability) IMO. Over time I'd like to change that.

grekko:
> Not sure I understand what you mean by "put them inside the template code" above … exact same function.

I was just not aware that the jinja2 processing happens at the python layer and assumed it is handled by the javascript code.

> Also, more regarding translation - if you are suggesting moving the strings out of the JS …

Nope. That's not what I was aiming for.

> Regarding naming: I just want things consistent.

Same here. Thats why I followed the existing `snake_case` pattern. Its just pretty unusual for JavaScript code these days :)

grekko:
> This is in fact the biggest issue I have with the current organization of our JS code - it's broken up somewhat arbitrarily by "what kind of functionality is this" rather than being organized by "what is this code doing".

> So you almost always have to open up and keep track of multiple files when you want to understand something or make changes.

That is something I also observed. Looking forward to re-structuring that. If you already have any ideas I'd be happy to learn about them.

grekko:
Otherwise I think I adressed the points you made.

PR #1809, opened 2017-03-23
Prevent invalid tagging operations

BjarniRunar:
This PR prevents certain invalid or dangerous tagging operations in the default UI; things like manually adding Drafts or trying to tag messages that haven't been indexed yet. These ops could result in user confusion or even internal data corruption.

This PR mostly addresses the low level concerns, while also fixing part of the UI. The UI side will still need more work later on, in particular it's not immediately obvious when dragging/dropping of tags or messages is expected to fail.

### DIFF
--- a/mailpile/plugins/tags.py
+++ b/mailpile/plugins/tags.py
@@ -37,6 +37,8 @@ _plugins.register_config_section('tags', ["Tags", {
     'flag_hides': ['Hide tagged messages from searches?', 'bool', False],
     'flag_editable': ['Mark tagged messages as editable?', 'bool', False],
     'flag_msg_only': ['Never apply to entire conversations', 'bool', False],
+    'flag_allow_add': ['Allow users to apply this tag', 'bool', True],
+    'flag_allow_del': ['Allow users to remove this tag', 'bool', True],
 
     # Tag display attributes for /in/tag or searching in:tag
     'template': ['Default tag display template', 'str', 'index'],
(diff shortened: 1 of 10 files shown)


mailpile/plugins/tags.py
```diff
@@ -361,29 +369,45 @@ def _do_tagging(self, ops, msg_ids, conversations, save=True, auto=False):
...
                                                           default=[]):
                         idx.add_tag(self.session, t._key, msg_idxs=msg_ids)
             else:
                 self.session.ui.warning('Unknown tag: %s' % op)
 
+
+        if rv['ignored'] and (len(rv['tagged']) == len(rv['untagged']) == 0):
+            self.event.private_data['ignored'] = rv['ignored']
+            return self._error(_('Nothing Happened'), result=rv)
+
```
karlp: Is it worth logging at debug or info that _some_ actions were ignored based on these flags as well?  not just when you ignored everything?

karlp:
reads generally sanely fwiw.

mailpile/plugins/tags.py
BjarniRunar: I struggled with that a bit; I agree this would be a useful enhancement for the future.

In practice this is a pretty rare event and the UI behaviour will actually be relatively sane: This only comes up (AFAICT) if the user is viewing a compound result (so a union of Tag A and Tag B), where Tag A is removable but Tag B is not. If the user then clicks the "untag" button or drags the message somewhere else, the result at the backend is that the removable tag (Tag A) will be stripped off the messages (but not Tag B), and the result in the user interface is the messages will disappear from view because they're no longer in the union.

As you point out, the user won't be informed that Tag B wasn't stripped, but everything else will behave correctly. In general, explaining to the user in the UI that some tags are removable but others are not is a hard problem and I could use some help from someone with visual design skills there.

karlp: leave it for later then.  Was just something I thought of reading the diff.  the way some other things do it is changing the cursor when the drop target is illegal, but that's for "all or nothing" cases again

PR #1810, opened 2017-03-23
Fix problems with old PGP-encrypted metadata indexes

BjarniRunar:
This PR does two things:

   1. Brings back the ability to load a PGP-encrypted metadata index
   2. Disables generation of (new) PGP-encrypted metadata unless we have no master secret


karlp:
:+1: works for me to startup again.

BjarniRunar:
@karlp: thank you for testing and reporting back! Merging.

PR #1815, opened 2017-03-28
[wip] Expose some debugging method to web ui

grekko:
This is a WIP :)

The goal is two-fold:

1. Re-enable the search bar to allow handling `special queries`. Currently the search bar only allows searching e-mails. As a result we would be able to use `in:inbox` or `in:sent` to navigate mailboxes via the keyboard + searchbar.

2. Implement a first `special query` to give users access to debugging info via `cli queries` through the web UI. A first proof of concept would be the `/print sys.env` which should print the current settings.

TODOs
-------
- [x] Re-enable search bar
- [ ] Provide cli query endpoint for cli commands like `/print sys.env`
- [ ] Implemenent web UI handling for displaying the result of `/print sys.env`


grekko:
@BjarniRunar Can you have a look at this and maybe tackle the TODO for providing the cli query endpoint on the python side?

shared-data/contrib/autoajax/autoajax.js
```diff
@@ -133,15 +133,6 @@ prepare_new_content = function(selector) {
             });
         }
     });
-    var $form = $(selector).find('form#form-search');
```
grekko: autoajax used to handle all search queries to the form. Now those are processed again by the [code in the topbar.js](https://github.com/mailpile/Mailpile/pull/1815/files#diff-146d89056853750bae4bf5d688a842c3R35).

That code will forward all non special-queries back to autoajax.

shared-data/default-theme/html/jsapi/index.js
```diff
@@ -400,6 +400,7 @@ Mailpile.API = {
 
 
 // JS App Files /=============================================================
+{% include("jsapi/global/polyfills.js") %}
```
grekko: To be able to use `modern` JavaScript methods on the String.prototype, e.g. `"x".startsWith("y")` or `"x".includes("y")` I included those polyfills for IE browsers which do not support either [startsWith](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes) nor [includes](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes).

Another solution to manually managing those might be to use something like https://github.com/inexorabletash/polyfill

shared-data/default-theme/html/jsapi/global/global.js
```diff
@@ -5,38 +5,12 @@ Mailpile.fix_url = function(url) {
   return url;
 }
 
-Mailpile.go = function(url) {
```
grekko: This method has been [monkey-patched](https://github.com/grekko/Mailpile/blob/4c0384fbfed1d44bd7761818176822c93e1fd79e/shared-data/contrib/autoajax/autoajax.js#L234) by autoajax and therefore never been executed again.

shared-data/default-theme/html/jsapi/index.js
BjarniRunar: This is fine! :+1:

shared-data/default-theme/html/jsapi/global/global.js
BjarniRunar: You are missing the idea that autoajax is, technically, optional. Please back these changes out so things continue to work without autoajax. :-)

shared-data/contrib/autoajax/autoajax.js
BjarniRunar: Autoajax only interferes on submit. This means that if autoajax is not present (it is an optional thing at the moment), a normal submit happens and everything still works. Treating autoajax as optional forces certain coding practices which increase the odds that things will actually have a chance of working with Javascript completely disabled. So please leave that mechanism intact!

shared-data/default-theme/html/jsapi/ui/topbar.js
```diff
@@ -31,44 +31,60 @@ $(document).on('click', '#form-search .clear-search', function(e) {
   $(e.target).hide();
 });
 
+/* Search - Special handling of certain queries */
+$(document).on('submit', '#form-search', function(event) {
+  var commands = {
+    in: function(cmdArgs) {
```
BjarniRunar: This is unnecessary - in:inbox is a search term and all /in/tagname/ does is invoke the search function with in:tagname as a parameter. The only difference is the URL in the location bar, which isn't worth adding the maintenance burden of trying to figure out which tags exist at this point in the Javascript. :)

BjarniRunar:
I'm going to suggest we discuss this a bit more!

One of the main questions is: Do we want this new functionality to work without Javascript? If yes, why? If no, why not? The criteria we have used until now has been that "read only things work without Javascript", and that includes all searches and a bunch of other things that the CLI allows.

If it is to work without Javascript (which I think might be desirable), then the work should start at the Python layer and then once we have something working we check whether there's any progressive enhancements to be gotten from the world of Javascript that we weren't already getting for free using Autoajax.

Issue #1811, opened 2017-03-24
Attachments: Prevent aborted/partial uploads


BjarniRunar:
Currently, the Mailpile UI will allow the user to navigate away from the page where an attachment upload is still in progress.

We prevented premature sending in #1656, but that's only a fix for the most embarrassing failure mode (where a message is sent without its attachment), there are quite a few other ways where this can go wrong:

1. In cases where the navigation is "autoajax" enhanced (partial page updates), this doesn't actually interrupt the upload but there remains no user-visible indication that an upload is still in progress. The progress bar from #1646 should probably be moved to a notification element so it persists even if the user navigates away (see also issue #1813).

2. In the case where the user triggers a full-page refresh, the upload itself will be aborted. We need to detect that this is about to happen, warn the user and give them the choice whether to continue (and abort the upload) or wait a bit longer.

The attachment Javascript logic is (for the most part) in `shared-data/default-theme/html/jsapi/compose/attachments.js`.

BjarniRunar:
Note: Putting this issue in the 1.0 milestone - but it's a "nice to have", I don't consider it a blocking issue for a release.

grekko:
> In cases where the navigation is "autoajax" enhanced (partial page updates), this doesn't actually interrupt the upload but there remains no user-visible indication that an upload is still in progress.

I tried to validate that but noticed that the upload is actually being canceled, at least in Google Chrome 57 on Mac OS X.

![](https://giant.gfycat.com/FlamboyantAltruisticGemsbuck.gif)

I checked the current behavior of Gmail for "navigating away" while uploading attachments. For Gmail the XHR upload does not cancel and I can navigate back to my E-Mail draft and after the XHR upload is done, I can successfully access the file/send the email.

Although I noticed that while the XHR upload is still in progress, Gmail does not re-open the Compose view for the E-Mail draft (which still has a pending upload). Not sure if thats a bug/feature.

@BjarniRunar Can you verify that the upload is indeed canceled or describe how to keep the upload "alive" while navigation to a different view?

grekko:
I dont know how important it is for people to be able to leave the compose page while uploading attachments but I would argue that it is OK to force people to stay on the current page and wait for the upload to finish.

What would be missing right now is some UI feedback for the user that informs about the fact that leaving the site would cancel the upload and asks her wether she wants to continue and cancel or wait for the upload to finish.

BjarniRunar:
[...]

PR #1821, opened 2017-04-08
Implement `onbeforeunload`-event handler for attachment uploads #1811

grekko:
Partially adresses #1811 

If a user uploads an attachment and triggers a full-page-refresh we want to notify him about in-progress uploads and the fact that the full-page-refresh (or unloading the document) will cancel the upload.

The `unbeforeunload`-event can prompt the user with either our custom or a generic message (depending on the user agents implementation) wether she either wants to continue with the full-page-refresh or wants to wait for the page to "finish it works".

### DIFF
--- a/shared-data/default-theme/html/jsapi/compose/attachments.js
+++ b/shared-data/default-theme/html/jsapi/compose/attachments.js
@@ -99,8 +99,11 @@ Mailpile.Composer.Attachments.UpdatePreviews = function(attachments, mid, file) 
 };
 
 
-Mailpile.Composer.Attachments.Uploader = function(settings) {
+Mailpile.Composer.Attachments.Uploader = {
+  instance: false
+};
 
+Mailpile.Composer.Attachments.Uploader.init = function(settings) {
   var uploader = new plupload.Uploader({
   	runtimes : 'html5',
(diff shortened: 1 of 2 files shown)


shared-data/default-theme/html/jsapi/compose/attachments.js
```diff
@@ -1,5 +1,16 @@
 /* Compose - Attachments */
 
+// Prompts the user if a full-page-refresh happens while uploading attachments.
+window.addEventListener("beforeunload", function (event) {
+  if (Mailpile.Composer.Attachments.Uploader.hasPendingUploads()) {
+    var confirmationMessage = "There is still an attachment upload in progress. Are you sure you want to cancel the upload?";
```
BjarniRunar: This needs i18n :-)

grekko: Good catch. Added with 3d08b6a :)

BjarniRunar:
This is great. Thank you for reformatting it so I could easily track the changes!

The only nit-pick I have, is the current implementation hard-codes in the assumption that there is only ever one upload in progress at a time, which may not always hold true. But we can address that later I think, this is definitely a step in the right direction.

PR #1824, opened 2017-04-10
Enable e-mail deletion

BjarniRunar:
This is a big, scary PR which does two important things:

1. Enable actual deletion of e-mails, both locally and from remote servers
2. Add a "hints" mechanism which lets us explain aspects of the app via. notifications that show up after the user has been using Mailpile for a while.

These are together, because I don't want to ship 1. without 2. - this first hints talk about deletion and backups.

I've done my best to break all the scary bits of 1. into bite-sized individual commits to facilitate review.


BjarniRunar:
Merging this; looks like nobody has time to review it. It needs to be tested in the wild...

PR #1827, opened 2017-04-13
Set execution bit for dockers entrypoint

tomaspapan:
Without this fix, docker run will fail by not executing entrypoint.sh
because of missing permissions (missing execution bit).


BjarniRunar:
Thank you!

PR #1832, opened 2017-04-20
Fix encrypted attachment filename handling and encoding.

### LINKED ISSUE #1789: Downloading PGP-encrypted content buglets
Mailpile will currently transparently decrypt any attached PGP-encrypted content, which is nice.

However, this feature has two problems that need fixing:

   * [ ] A way to download the encrypted data (w/o decrypting)
   * [x] The decrypted data is served up with the wrong filename.

The latter issue is particularly vexing and needs to be fixed before 1.0.

Why it's bad: you download a file named `foo.zip.gpg`, which should actually be named `foo.zip`! Since it's named .gpg, the natural (or automatic) reaction is to open it using GnuPG which doesn't know what to do with the contents, which will cause even experienced users to be confused (e.g. assuming the data was corrupted somehow). Only a lucky few will think to try and directly unzip a file named `foo.zip.gpg`...

JackDca:
Fixes #1789

### DIFF
--- a/mailpile/crypto/mime.py
+++ b/mailpile/crypto/mime.py
@@ -74,40 +74,28 @@ def _update_text_payload(part, payload, charsets=None):
 ##[ Methods for unwrapping encrypted parts ]###################################
 
 
-def MimeReplaceFilename(header, filename):
+def MimeAttachmentDisposition(part, kind, newpart):
     """
-    Accepts a MIME header containing filename="...".
-    Replaces it with filename.
+    Create a Content-Disposition header for a processed attachment using the
+    original file name if available from the PGP packet, otherwise try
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Wow, great work.  This looks great! I will merge as soon as I've had a moment to quickly test it myself.

JackDca:
Thanks! I was not able to find email software that created token (i.e. unquoted) filename parameters.  I tested that aspect by manually modifying some emails in an mbox file.  From your comments on #1789 you obviously receive some emails with token filenames. So it would be good if your quick test could include some of those emails.

JackDca:
Neat solution with unwrap_attachments!

PR #1833, opened 2017-04-25
Front end chore: Remove dead code, adjust String prototype polyfill

grekko:
While browsing through the front end code I discovered some unused js helper code.

Also I modified the provided polyfill to `String.prototype.startsWith` to be more compliant with the [ECMAScript definition](http://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.startswith).


BjarniRunar:
Thank you!

PR #1834, opened 2017-04-26
This PR enables TLS certificate verification, adds tools to manage TLS certs

BjarniRunar:
See individual commits. :)


JocelynDelalande:
Just tested, I like it :-)

However, I feel like « add an exception » is quite vague, it feels like « bypass all the securities ». In fact this exception thing is (to my understanding) more like « trust *this* certificate on next uses ».

Just a user thought/feeling, wanted to write it somewhere :-)

Issue #1836, opened 2017-05-01
Long From Address causes overlap making it hard to read


jmcclelland:
These screen shots were made while running Firefox 52.0.2-1 on Debian Stretch.

A long from address that is perfectly normal causes a bit of text to be cut off:

![normal](https://cloud.githubusercontent.com/assets/4511942/25581445/e425de5a-2e55-11e7-99c7-88e4fd8c4090.png)

That's bearable. However, a really long from address causes overlapping making it unreadable.

![long](https://cloud.githubusercontent.com/assets/4511942/25581447/e9190e78-2e55-11e7-81a9-b08cbb3299b1.png)

jmcclelland:
The pushes is an improvement but not ideal since it loses the flow of the from address around the icon. Hopefully someone can improve on it.

![fixed-normal](https://cloud.githubusercontent.com/assets/4511942/25581573/ac8df26a-2e56-11e7-9f39-542d1307af77.png)

![fixed-long](https://cloud.githubusercontent.com/assets/4511942/25581572/a8af53b4-2e56-11e7-9239-a8da7536ee2e.png)

BjarniRunar:
I think the simplest solution here is to prevent wrapping and set overflow: hidden on the container. There should be tooltips that show the full content when it gets truncated.

PR #1837, opened 2017-05-01
avoid overlap in display of from address

jmcclelland:
This fixes the problem, but is not ideal

https://github.com/mailpile/Mailpile/issues/1836


BjarniRunar:
Hi! Thanks for doing this. I merged, but then had to revert because this was going to get lost anyway.

It's fine to edit default.css directly, but you also need to edit the .less source files or your changes will get lost next time we rebuild CSS file from source.

If you can be bothered, it would be great if you could file another PR - and maybe include a screenshot so we can see what things look like after the change?

PR #1852, opened 2017-06-23
Use AddressHeaderParser instead of ExtractEmails (fixes #1819)

### LINKED ISSUE #1819: Buggy address parsing
There is an issue with the way Mailpile parses `To:` `Cc:` addresses.

With 

    To: Foo <foo@example.com>

It will try to send to foo@example.com. That's fine :-)

With 

    To: foo@bar <foo@example.com> 

MP will try to send to two mail addresses : 

- foo@bar 
- foo@example.com

Which seems wrong, because `foo@bar` is the name of the contact, not its email.

(@BjarniRunar if you give me some hints about where to dig, I could try to tackle this issue)

smari:


BjarniRunar:
Awesome, thank you.

Issue #1730, opened 2016-12-16
Setup fails when Mailpile detects insufficient disk space


BjarniRunar:
The setup process fails to create the default tags and such things if Mailpile detects that it doesn't have enough disk space to download mail. This makes the app launch in a weird wonky state that it won't recover from automatically, even if the user frees up disk space.

Proposed fix: The setup process (which lives in `mailpile/plugins/setup_magic.py`) should temporarily set `sys.minfree_mb` to a very low value, and reset it back to the default when it finishes.

Workaround: If you're bitten by this, the following steps on the Mailpile CLI will fix it;

    mailpile> set sys.minfree_mb = 10
    ...
    mailpile> setup
    ...
    mailpile> unset sys.minfree_mb

(or if you're sure you want to run Mailpile with less than 1GB of free space for mail and indexes, set `sys.minfree_mb` to whatever value you are comfortable with instead of unsetting it.)

MikeinRealLife:
I will attempt to do this.

BjarniRunar:
I've verified this is fixed. Thanks @smari !

PR #1855, opened 2017-06-23
Less ExtractEmails, more setup success on low disk space environments.

smari:
Does more towards #1852 (which is already closed though), and also deals with #1730.


BjarniRunar:
Excellent, thank you!

PR #1868, opened 2017-07-13
Lots of mobile/responsive updates.

smari:
This brings us significantly closer to responsive UI. There are some known things that are missing still, in particular the profiles index. I'm working on that, but want to suggest merging at this point in order to not drift too far from HEAD and keep progress public.


BjarniRunar:
Takk!

PR #1891, opened 2017-07-25
fixed layout for pile table  (removes horizontal scroll)

JocelynDelalande:
Horizontal in mailpile have been annoying me for a while, so I decided to dig that issue :-).

After hours of struggling, experimenting and reading, I figured out that the "auto" `<table>` sizing makes it quite impossible to prevent the table from overflowing. Unless you know what your content looks like by advance (which is not the case with mailpile, for example mail subjects with very long words mess up table sizing).

For the record, #1765 seemed to solve some scrollbar issues, and introduced others.

See detailed commit messages for details.

PS: I tried several approaches. One of them included to "escape" the unfolded message line from the main table, so that it do not share the columns widths with folded messages (different data, different column set), but I figured out a simpler solution. I keep that idea of having different column size for a later PR, maybe. But I already know that @BjarniRunar is attached to the same-sized columns among message summary and expanded message.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -8380,7 +8380,7 @@ tr.modal-tag-picker-item td.checkbox {
   }
 }
 #pile-results td.date {
-  width: 60px;
+  width: 80px;
   text-align: right;
   white-space: nowrap;
   color: #b3b3b3;
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -8237,8 +8237,7 @@ tr.modal-tag-picker-item td.checkbox {
   cursor: move;
 }
 #pile-results td.crypto-and-tags,
-#pile-results td.message-nav,
-#pile-results td.avatar {
+#pile-results td.message-nav {
   width: 24px;
   text-align: center;
   white-space: nowrap;
@@ -8259,6 +8258,18 @@ tr.modal-tag-picker-item td.checkbox {
 #pile-results.snug td.avatar {
   padding-top: 4px;
   padding-bottom: 3px;
+}
+#pile-results.comfy td.avatar {
+  width: 32px;
+  padding-right: 8px;
+}
+#pile-results.cozy td.avatar {
+  width: 24px;
+  padding-right: 6px;
+}
+#pile-results.snug td.avatar {
+  width: 18px;
+  padding-right: 4px;
 }
 #pile-results.comfy td.avatar a img {
   width: 24px;
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -7,15 +7,15 @@
 ================================================== */
 /* @license
  * Mailpile font designed for Mailpile
- * 
+ *
  * You may obtain a copy of the license at the URLs below.
- * 
+ *
  * Webfont: Mailpile by Brennan Novak
  * URL: http://github.com/mailpile/fonts
- * 
+ *
  * Mailpile font is licensed under the SIL Open Font License (OFL)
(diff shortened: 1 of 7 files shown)


BjarniRunar:
This is promising, but needs a little bit more work. Issues I see at the moment (see screenshot):

- Ugly lack of padding to the right of the avatar
- Layout cannot handle multilpe tags; they overflow into/under the subject

The latter is tricky; I'm not sure what the solution is. Perhaps getting rid of the tags column altogether and putting the tags into the From table cell with "float: right" would work?

screenshot from 2017-07-26 11-59-33

JocelynDelalande:
> his is promising, but needs a little bit more work. Issues I see at the moment (see screenshot):
> 
>  Ugly lack of padding to the right of the avatar

Ok. Actually, I made my tests with "cozy" display style, while you are using "comfy" one. That's why. I'm going to rework and test with cozy/comfy/snug styles.

>  Layout cannot handle multilpe tags; they overflow into/under the subject
> 
> The latter is tricky; I'm not sure what the solution is. Perhaps getting rid of the tags column altogether and putting the tags into the From table cell with "float: right" would work?

That would be a solution. That let us with 3 solutions :

1. squat the right side of the sender column (the solution you suggested)
2. Display tags at the beggining of the subject column (as gmail does), zero-overlap guaranted
3. Increase the space dedicated to tags (but that is a loss of space for people not using tags at all)

@BjarniRunar  I'm inclined towards 2. What do you think ?

BjarniRunar:
I prefer 1 to 2. Going with 2 means we'll end up with a jagged edge where text begins at varying indents which I find very ugly, whereas 1 is just using the same space as we are now - the only difference will be that instead of the tags being centered, they will be right justified.

Solution 3 isn't a solution IMO, because we cannot predict how many tags people have and we're not going to dedicated ALL THE SPACE to tags... It's exactly what we have now, it just breaks at a different number of tags.

JocelynDelalande:
So first, the avatar padding/sizing issue is solved :-).

Then, I am not sure to agree on the uglyness of subjects started by tags (I don't find it ugly at gmail). But, I'm not inclined into bikesheding, so I will give a try at *1.* `. A bit later today, maybe, or tomorow.

JocelynDelalande:
TODO

- [x] fix avatar padding
- [ ] (WIP) include tags in `from` column, right-floating.
- [ ] fix distorded avatar when a message is unfolded

JocelynDelalande:
I think I adressed identified issues on this PR :-).

Some notes:

There is no possible layout that can handle nicely an unlimited amount of tags. In case there is too many tags for the available space, it will behave like that : 

image

Which is not pretty, but better than horizontal scrolling.

IMHO, this case is quite rare, and could be addressed with some JS doing an ellipsis of some tags.

The display of tags for the unfolded message was weird (even before this patch). So I moved them to a better (IMHO) place:

image

## Future

I think the case shown on the first screenshot could be handled nicely with an ellipsis of the overflowing tags with some nice "show all in popup". But that might imply a bit of JS and is out of the scope of this PR.

**What do yo think ? Feedbacks/tests more than welcome :-)**

JocelynDelalande:
bump. The layout on master branch is still scrolling horizontally at my screen size ;-).

JackDca:
Tried this on my tablet - Asus TF700T - Android 4.4.4 (Cyanogenmod) - Chromium. It fixes the missing handle problem so emails can be dragged to a tag using the touchscreen interface. Using the touchscreen is still a bit difficult but that's another issue. Thanks for pointing me at this PR JocelynD!

JocelynDelalande:
@BjarniRunar any feedback ?

BjarniRunar:
Sorry about the really late response - I finally got around to testing this, and AFAICT, it's great. Merging!

Issue #1599, opened 2016-05-02
Improve Mailpile's responsive rendering


pchampin:
When trying https://demo.mailpile.is/ in my mobile (or on my laptop and reducing the size of the browser window), I notice that some effort is made to make the application responsive. However, I find it unsusable on my Android smartphone (with Firefox and Chrome alike), because everything is too small.

This is due to the fact that smartphone are lying about their resolution, in order to correctly display non-responsive websites. However, a propertly responsive application should inform the smartphone _not_ to lie about it.

Although the future standard is [not yet supported](http://caniuse.com/#search=%40viewport), the current de-facto solution, namely [&lt;meta name="viewport"&gt;](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag?redirectlocale=en-US&redirectslug=Mobile%2FViewport_meta_tag) seems to work in many situations.

This would be a very nice improvement.

broder:
Seems to have been added here: https://github.com/mailpile/Mailpile/blob/74d1eea99bdb892555b0cf191b612b9340df1c48/shared-data/default-theme/html/partials/head.html#L3

jamestomasino:
Some quick testing makes it look like there will be some follow-up work on this change. Once the viewport meta tag is in place, the mobile styles for the default theme are wider than the device.

![mobile-styles](https://cloud.githubusercontent.com/assets/201422/23324417/a8041360-fabb-11e6-9564-4100b69e667e.jpg)

BjarniRunar:
Yes, more work is needed. Renamed the issue to reflect that!

grekko:
@BjarniRunar Do you have any specific goals already on the top of your head?

I shortly checked some pages with a limited viewport. I decided to test it with 320 CSS pixels width (e.g. the iPhone 5 comes with that) which seems to be the lower boundary for most known devices according to https://mydevice.io/devices/

I checked some pages, e.g. the `/profiles/` url and most actions thing seem to be accessible. The layout is a little bit funny in some places but that does not prevent me from accessing the page.

![The `/profiles/`-url with 320 CSS pixels width](https://dl.dropboxusercontent.com/s/q3pfytq6337rdut/2017-04-25%20at%2012.39%20%281%29.png)

## Some things I noticed

[...]

PR #1892, opened 2017-07-25
Enhance footer layout on smartphone

JocelynDelalande:
Using existing mediaqueries.

I think the speedometer is a nice gadget, but it does not worth the screen space it takes on a smartphone display.

## Before

image

## After

image

Relates #1868 

@smari as you pushed most of the other responsive stuff, what do you think of this one ?

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -9926,7 +9926,8 @@ a.modal-tag-color-option:hover {
   }
   /* Stuff we just hide... */
   #sidebar hr,
-  #notifications-header {
+  #notifications-header,
+  #pile-speed {
     display: none;
   }
   #pile-results,
@@ -10051,10 +10052,6 @@ a.modal-tag-color-option:hover {
   }
   div.footer-nav {
     display: none;
-  }
-  #pile-speed {
-    margin-bottom: 0px;
-    font-size: 70%;
   }
   #display-density-selector {
     display: none;
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -10059,6 +10059,18 @@ a.modal-tag-color-option:hover {
   #pile-more {
     display: none;
   }
+  #pile-bottom .button-primary {
+    width: 49.3%;
+    float: none;
+    text-align: center;
+    margin-left: auto;
+    margin-right: auto;
+  }
+  /* Single  footer button case */
+  #pile-previous.button-primary:last-of-type,
+  #pile-next.button-primary:nth-of-type(2) {
+    width: 100%;
+  }
   table.account-list {
     margin: 0 0 !important;
     display: block;
--- a/shared-data/default-theme/less/app/mobile.less
+++ b/shared-data/default-theme/less/app/mobile.less
@@ -278,7 +278,8 @@
 
   /* Stuff we just hide... */
   #sidebar hr,
-  #notifications-header {
+  #notifications-header,
+  #pile-speed{
     display: none;
   }
 
@@ -359,11 +360,6 @@
 
   div.footer-nav {
     display: none;
-  }
-
-  #pile-speed {
-    margin-bottom: 0px;
-    font-size: 70%;
   }
 
   #display-density-selector {
--- a/shared-data/default-theme/less/app/mobile.less
+++ b/shared-data/default-theme/less/app/mobile.less
@@ -370,6 +370,20 @@
       display: none;
   }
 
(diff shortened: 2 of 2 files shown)


BjarniRunar:
```
(11:47:39) BjarniRunar: JocelynD: your big centered Next button..
    what happens when there's also a Previous?
(11:47:50) JocelynD: BjarniRunar: uh oh :-)
(11:47:55) BjarniRunar: hehe
(11:47:57) JocelynD: BjarniRunar: please leave a coment on PR :)
```

JocelynDelalande:
You are damn right. Fixed that major case I forgot, see complementary screenshot:

image

BjarniRunar:
Thank you!

Kabouik:
After I finally installed Mailpile on a server, I can access it from my smartphone. In my opinion there is still too much UI stuff and not enough content to make it comfortable to use. Seeing just 2.5 truncated rows in Inbox view brings frustration when checking emails I think.

The above pictures from @JocelynDelalande show it in portrait mode, and the below pictures show it in landscape mode on a 5" screen (this is a Sailfish phone, but I've tried with an Android browser and it's basically the same except for the URL bar being at the top):



This is the minimal and default zoom level and I think it's just not enough. Since it's both the minimum and default, the only thing is zoom in, but it's not really useful or relevant: 



Overall, I think there are a few key points where space could be gained because it basically does not bring any improved ergonomy:



1. Buttons on the left are convenient because they are big and easy to select with thumb, but there is still some significant space loss. I think overall the grey bar could be narrower by reducing margins around icons (and margins between icons and their text). For reference, the buttons in the URL bar of my browser are totally big enough for thumbs; they still have big margins but they don't actually need them, see for instance the URL bar in another native browser with smaller margins (and still no issue with using thumbs on this buttons):
image

2. Rows need to be high enough to be easy to select, but I think it's still too big for what is the minimal zoom level. It shows only a total of 3.5 rows on a 5" screen. Row height could probably be reduced to the height of the contact icons on the left and still be easy to select. The Display snug/cozy/comfy options do not do much on mobile view. This is snug here.

3. The information row about selection checkboxes doesn't seem critical to take this much space. Wouldn't users just realize that they can select emails by taping on them without this information? Do we really need a "Select all" in mobile view? What about removing this row, but moving individual checkboxes onto the corner of contact icons? This would make it obvious that emails can be selected, and this would also indicate where to tap to select (Mailpile doesn't clearly indicate areas where to click to select/open/search by contact).

4. There is lost space here, at least in landscape mode. This is less obvious in portrait but the grey bar height is still huge (and to a lesser extent, the height of the search field seems a bit big as well). Do we need it to be displayed all the time? What about showing it just when at the top of Mailpile view, but hiding it when scrolling down? 

5. There is lost space as well, and actually the left icons are search icons. What about moving them to the  top bar next to the search field? This would make a consistent "Search" bar with all search tools in the same place, and Order and Display could be grouped together in the top right (and still shown only when at the top of the window, it's not necessary when scrolling down). This would work best in landscape mode, but would be harder to achieve in landscape of course (unless all search icons could be all grouped in a single hamburger icon?).

6. After playing a bit with the search field on mobile view, I noticed it was not big enough to type in an email address, at least at the current minimal zoom level. This means that if you want to edit a search or add search terms, you have to clear it beforehand with the "X" button on the right. You may hit this button anyway if you try to place the cursor at the end of your search terms. A wider field would help, or just smaller font, or a combination of both, with just enough space on the right to place the search icons and order/display menu. In landscape, there's a lot of space for whatever you'd like. In portrait, maybe the bar could just be Mailpile icon, Search field (without the blue button, who would use it instead of just hitting return on the keyboard?), Search icons grouped in a hamburger icon, Order/Display menu. I think it would fit even in portrait mode. Probably nobody would use any of these tools on a smartphone when scrolling so it'd be fine if they only show up when at the top of the page.

These are just ideas/suggestions, and I figured I'd just write them down before they go away (I don't think I will use Mailpile mobile view extensively before it can show much content, so I could forget quickly), but of course you do whatever you want with them, including burning and throwing to the trash. I do know that the mobile UI is work in progress and that at the moment there is no real mobile vs desktop views, but rather resolution-dependent layouts, and therefore all the above suggestions would be *a lot* of work to implement, but again this is just a list of improvement ideas you may agree with or not; not demands or urgent requests or anything like that. It's already great to see that most function do work on a mobile browser actually!

JocelynDelalande:
@Kabouik thanks for that long post ; could you please copy it in #1599 so that we don't loose it in this closed PR :-).

Kabouik:
Sure, done!

PR #1897, opened 2017-07-28
Guess typical folder names on servers independent of locale

FlakyBit:
When setting up Email accounts on servers that use different language than is setup in Mailpile, try to correctly associate tags with the (translated) remote folder names.


BjarniRunar:
Takk!

Issue #1896, opened 2017-07-27
Double-check attachment behaviour: avoid HTML, SVG, JS phishing


BjarniRunar:
HTML, SVG and JS all pose serious security/phishing risks if opened by the browser directly from e-mail. We need to double-check to make sure such attachments are disposition: attachment and trigger the save dialog, not the inline rendering.

PR #1898, opened 2017-07-28
Only give browser inline hint for safe mime types (#1896)

FlakyBit:


BjarniRunar:
Thank you!

BjarniRunar:
This should resolve #1896

PR #1924, opened 2017-08-27
Use .text to print doc objectify.StringElement

irregulator:
Autoconfig's xml documentation element may contain localization
strings [1]. If these strings contain non-ascii characters, e.g.
greek, autoconfig xml parsing will produce UnicodeEncodeError
exception and mailpile's account setup autoconfiguration will
fail.

This is due to the string casting of the objectify.StringElement.
This commit uses the .text method to print either type str or type
unicode.

[1]: https://autoconfig.thunderbird.net/v1.1/freenet.de

### DIFF
--- a/mailpile/plugins/setup_magic.py
+++ b/mailpile/plugins/setup_magic.py
@@ -575,7 +575,7 @@ class SetupGetEmailSettings(TestableWebbable):
                         result['docs'] = result.get('docs', [])
                         result['docs'].append({
                             'url': docs.get('url', ''),
-                            'description': str(docs.descr)
+                            'description': docs.descr.text
                         })
                 except AttributeError:
                     pass


irregulator:
This is pretty easy to reproduce.

Suppose example.xml:
 
```
<?xml version="1.0" encoding="UTF-8"?>

<clientConfig version="1.1">
  <emailProvider id="provider.gr">
      <documentation url="https://provider.gr/guide">
      <descr lang="el">Οδηγίες χρήσης</descr>
      <descr lang="en">Usage instructions</descr>
    </documentation>
  </emailProvider>
</clientConfig>

```

then:

```
with open('/tmp/example.xml', 'r') as f:
    xml_data = f.read()
    data = objectify.fromstring(xml_data)

for docs in my_data.emailProvider.documentation:
    print(str(docs.descr))
```
produces: `UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-6: ordinal not in range(128)`

BjarniRunar:
Thank you for this! I will review and merge as soon as I have a bit of time.

mailpile/plugins/setup_magic.py
```diff
@@ -577,7 +577,7 @@ def _get_xml_autoconfig(self, url, domain, email):
                             'url': docs.get('url', ''),
                             'description': str(docs.descr)
                         })
-                except AttributeError:
+                except (AttributeError, UnicodeEncodeError):
```
Rondom: Are you sure this is the right way to deal with this? The example XML you provide defines that it is encoded as UTF-8. Thus it should be processed according to this. Given that we can see the characters in your message here, they are valid UTF-8.

irregulator: You're actually right, I figured there is a better approach for this case. I'll update this pull request soon.

irregulator:
So, I followed the lxml's advice, https://github.com/lxml/lxml/blob/master/src/lxml/objectify.pyx#L728 , and used the .text method to print the documentation description StringElement. I think this is a better approach. 

Current code will simply keep the first discovered description even if more than one exist. Maybe we could handle this too, but I would prefer to keep this PR simple and focus on fixing the autoconfiguration functionality which is currently broken when parsing xml with unicode characters.

Cheers

Issue #710, opened 2014-06-11
Improve PGP compatibility of outoing mail?


BjarniRunar:
Currently, all PGP signed or encrypted messages we generate, use PGP/MIME. This is the most elegant and standards-compliant way to do PGP.

However, the real world is not elegant. This approach means those messages will not be verifiable or decryptable by people using tools like Google's new browser plug-in, that just operates on visible text on a web-mail site (the plug-in almost certainly has no access to the MIME structure of the underlying message). We had been assuming that such plug-ins were a dead-end and dying technology, but Google's entrance into the space means we should probably reconsider.

In the relatively common case where we are ONLY sending plain text, we could use in-the-body PGP ASCII armour encryption or signatures, instead of PGP/MIME. This will allow more people to read our messages. If we implement #632, this will not degrade our own user's experience at all.

Any reason not to do this, @smari , @brennannovak ?

BjarniRunar:
Folks using webmail or other clients which do not support PGP/MIME out of the box are a vast majority, not a minority. :-)

This was done a while back, haven't heard many complaints. Not sure why it is still open, really.

JocelynDelalande:
> Folks using webmail or other clients which do not support PGP/MIME out of the box are a vast majority, not a minority. :-)

I was speaking about gmail users only, because, other webmail/phone/heavy clients users don't have something like pgp/inline support at all and will see the noisy blobs. (this is particularly bad user experience on phones...).

BjarniRunar:
It may be an ugly experience, but it is also the only experience that allows them to copy/paste the blobs into an external tool and verify the signature or decrypt the data. The choice boils down to "ugly, but works" vs. "elegant but unusable".

Humans are good at ignoring stuff they don't understand, the --BEGIN PGP....-- preamble may be ugly but it does not prevent communication in my experience.

JocelynDelalande:
> It may be an ugly experience, but it is also the only experience that allows them to copy/paste the blobs into an external tool and verify the signature or decrypt the data. The choice boils down to "ugly, but works" vs. "elegant but unusable".

Makes sense.

Thinks like [webpg](https://addons.mozilla.org/en-US/firefox/addon/webpg-firefox/?src=search) goes in the direction of your point as well.

What about a default to only include armoured pgp to the people we know a key for ? (that means they are pgp-aware, and prevent the others from seeing the pgp blobs).

JocelynDelalande:
[...]

PR #1930, opened 2017-09-13
Expose « Prefer PGP/MIME » option in accounts UI

JocelynDelalande:


BjarniRunar:
I removed that checkbox from the UI on purpose... For me this is squarely in the category of "geeky things that geeky people can change on the CLI", not something that deserves space in the main UI.

As far as I can tell, this is not even a functional thing, it's bikeshedding - people want their e-mail formatted a certain way because they think it's more elegant, and this in spite of real world tests and feedback that show unambiguously that PGP/MIME is more confusing and less usable for non-technical recipients.

You're welcome to try to convince me otherwise, but this is one of the areas of the app that I've spent the most time thinking about...

JocelynDelalande:
> I removed that checkbox from the UI on purpose... For me this is squarely in the category of "geeky things that geeky people can change on the CLI", not something that deserves space in the main UI.

On this geeky console use-case ; with current code. If a geek changes this setting from CLI, that change will get lost if she happens to open the "account" UI and save (even without changing anything). That is confusing.

Furthermore the profile editing is currently not only for geeks, it is also painful (you have to targeta line number, a typo might destroy everything…). I know this is "for now" and it will improve. But as-is, even geeky users may get lost.

> As far as I can tell, this is not even a functional thing, it's bikeshedding
> people want their e-mail formatted a certain way because they think it's more elegant, and this in spite of real world tests and feedback that show unambiguously that PGP/MIME is more confusing and less usable for non-technical recipients.

I disagree, not to re-link the articles with pro/cons on the topic, I know you already read them, this is not a place to troll on that ;-). But a fact is that there is no clear consensus in favor of plaintext signatures.

But just to give my personal feedback : "my" newbie recipients are getting more confused (especially working with mailling-lists and mail forwards…) since I restarted my mailpile from scratch a few months ago and they started to send armored PGP signatures instead of the "unreadable attachment" they received before.

Another point is sending signed attachments to mailling-lists where some users are gpg users and other are not. pgp/mime offesr signature on attachments. This is valuable and legitimate to want if this use case is frequent (ie: that is my case).

So. I'm not contesting your default choice (especially #710 is a good argument). Just saying the preference for pgp/mime may be legit, and the choice up to power users if they want to. That is why « show me too many encryption settings » is here. For power users.

A strategic point which is worth mentioning is that a lot of current GPG users are using PGP/MIME and will appreciate to have this option. I know they are not first Mailpile target. But they will be 1st class Mailpile advisors, I think they deserve attention with that kind of details.

> You're welcome to try to convince me otherwise, but this is one of the areas of the app that I've spent the most time thinking about...

Well… Trying ;-). Good to see you around, anyway :-)

BjarniRunar:
Regarding attachments; if PGP/MIME is needed, it is used. It is used most of the time in fact, since it provides features that inline signatures and crypto do not.

The only time Mailpile (with its default settings) avoids PGP/MIME is if the message only has a single plain-text part. This is however a very common case, most e-mails have no attachments.

I agree that the profile editing is difficult.

And I know I'm going against the PGP-community's consensus by preferring inline signatures whenever possible; I am happy to be arrogant and assert they are collectively wrong about this. Much as they are wrong about things like the web of trust and a whole bunch of other things Mailpile is avoiding.

BjarniRunar:
I've decided to cave, and merge this. My goal is to then make the settings disappear again by somewhere adding an "expert mode" toggle that defaults to OFF.

PR #1933, opened 2017-09-23
Spelling: Only specify the folder

comradekingu:
Alternatively: Only specify the containing folder.
For people that don't know what a path is.


BjarniRunar:
Thank you! However, this time I disagree: Your proposed change doesn't actually capture what the problem is. I also think it's safe to assume that anyone that types ".." in a path knows what a path is. :)

comradekingu:
Translators might not,, and no amount of beard will help you then. :)

PR #1934, opened 2017-09-23
Spelling: Sendmail, …

comradekingu:
As per https://en.wikipedia.org/wiki/Sendmail


BjarniRunar:
Well, actually, the why not is "because it ads noise and work for the translators."

Still merging this, but at some point I'm going to start rejecting these just so things will stay still while the i18n work happens.

comradekingu:
Contrary to popular misconception, translators actually like to translate, and will happily improve upon a project they like.
Not only does changes keep people on their feet, it also gets eyeballs directed at other stuff meanwhile, which is a net benefit. The more you can engage with announcements and cooperation the better it gets.

PR #1936, opened 2017-09-23
Spelling: Poignantly pointy

comradekingu:


BjarniRunar:
Sorry about the very late response; but there are actually a whole bunch of things I disagree with in this one. I don't think it's good use of our time to debate them all, nor is it good use of translators' time to redo their translations.

However, your effort is not wasted. I am going through the whole list and manually fixing things at my end when I do agree they need fixing. Thank you for taking the time!

comradekingu:
@BjarniRunar I am just happy it happens :)

No translator has ever complained about "re-doing", as it were, it is a matter of clicking things from "history" on Transifex. Move to Weblate or Pottle and that is no longer even an issue, as they handle fuzzy strings correctly.

PR #1946, opened 2017-10-05
#1844: Search-bar Firefox font cutoff tweak

### LINKED ISSUE #1844: Font cut off below baseline in Firefox
This is a really minor issue. In Firefox characters having parts below baseline like like g and j  aren't displayed correctly in the searchbar, as shown below:

diff

I tried, but I couldn't find the source of the issue. Some CSS expert might be able to easily fix this.

I have a high-DPI screen, that could mess things up. It would be nice if someone could test this on a normal screen.

psyonara:
Fixes #1844.

Tweaked height and padding of search-bar so that Firefox doesn't cut off the bottom of letters like 'g' or 'j'. Tested in Chrome as well; if someone can confirm other major browsers, that would be great.

### DIFF
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -6915,8 +6915,8 @@ div.notification-bubble span.action {
 }
 .form-search input {
   width: 282px;
-  height: 18px;
-  padding: 9px 12px;
+  height: 20px;
+  padding: 9px 12px 7px 12px;
   float: left;
   font: normal 18px "HelveticaNeue", "Helvetica Neue", Helvetica, Arial, sans-serif;
   border: 0;
--- a/shared-data/default-theme/css/default.css
+++ b/shared-data/default-theme/css/default.css
@@ -6916,7 +6916,7 @@ div.notification-bubble span.action {
 .form-search input {
   width: 282px;
   height: 20px;
-  padding: 9px 12px 7px 12px;
+  padding: 8px 12px;
   float: left;
   font: normal 18px "HelveticaNeue", "Helvetica Neue", Helvetica, Arial, sans-serif;
   border: 0;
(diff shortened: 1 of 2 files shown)


JocelynDelalande:
Thanks for your contrib, this bug has been scratching me since ages :-)

That enhances a bit, but that is not enough. See screenshot :

screenshot-localhost 33411-2017-10-05-10-12-50


There seems to be something (which is IMHO, not only padding) "pushing" the text from the top. Good luck investigating :-).

(tested Fx 52.3.0 / Debian stretch)

psyonara:
Oh, that's a pity... Fx 56 seems fine. But yes, I see what you mean with the text being pushed down. I'll see if I can further tweak it.

psyonara:
I've tweaked the padding some more, and this is what it looks like in Fx 56.0 (Ubuntu 17.04):
search-padding
Any more tweaking and the font appears too far off centre (horizontally).

Can we call this "acceptable" in light of the issue being fixed on the latest (stable) version of Fx, or are we looking to cover multiple previous versions? It would seem that Fx has made some changes in how it renders the font, somewhere between v52 and v56.

JocelynDelalande:
@psyonara thanks :-).

image

Sadly, no improvement here with Debian / Fx 52.

> It would seem that Fx has made some changes in how it renders the font, somewhere between v52 and v56.

As we target providing debian packages for debian stable, I think this is consistent to provide something for the Fx version in Debian stable, WDYT.

btw, I'm getting suspicious about my Fx… could someone else using Debian stable test it ?

JocelynDelalande:
I upgraded my laptop, from a 1280x800 screen notebook to a 1400x900 one ; but the bug remains.

BjarniRunar:
I'm going to merge this - even if it doesn't work for everyone, it looks like a very reasonable change and from the comments it appears to help some of the time. Thank you!

PR #1948, opened 2017-10-09
Add missing lock on mail index

JocelynDelalande:
`update_ptrs_and_msgids()` seems to require consistent state of `self.INDEX`,
`self.MSGIDS` and `self.PTRS`. However, there is no lock to enforce it, neither in
`update_ptrs_and_msgids()` nor in its caller, `scan_mailbox()`.

Just wondering, that is the first time I read this part of the code. Is my analysis relevant ?


BjarniRunar:
I think you are right. Merging, thank you!

PR #1959, opened 2017-10-23
Fix crash of smtp_client when utf-8 received + rename last /usr/bin/python to /bin/env python2

katomaso:
After those two changes Mailpile works fine on my system

### DIFF
--- a/mailpile/plugins/vcard_mork.py
+++ b/mailpile/plugins/vcard_mork.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python2
 #coding:utf-8
 import sys
 import re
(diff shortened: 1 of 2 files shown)


JocelynDelalande:
@katomaso Hi, welcome around, and thanks for your contribution :-).

> Replace last python with python2

That is a good idea ; could you take the time to extend this PR to do the replacement for files containing a shebang lines pointing to */usr/bin/python* ?

(I'm curious, did you call *mailpile/plugins/vcard_mork.py* as a script to require this edit ?)

> Fix crash with real UTF-8 characters in SMTP 

Could you remove this edit please ? Exact same patch have been brought in by #1960

katomaso:
It will be my pleasure :-)

katomaso:
Closing this MR in favour of #1962

PR #1965, opened 2017-11-08
Remove excess whitespace from notification HTML

### LINKED ISSUE #1931: [Browser Client] Minor memory leak: Text nodes in notification area
## Problem

When Mailpile adds a notification to the notification area, it accidentally adds three DOM nodes instead of one. You can see this happening by examining `document.getElementById('notification-bubbles').childNodes` while navigating between individual emails and the inbox.

## Explanation

When mailpile adds a notification to the notification area, it produces a document fragment [using the following HTML template](https://github.com/mailpile/Mailpile/blob/b5a8526f60a1f7ea6f3aab30ac1a4bf467179492/shared-data/default-theme/html/jsapi/ui/notifications.js#L193):

```html
"
  <div id="event-<%= event_id %>" class="notification-bubble <%= status %> hide">
    <span class="icon <%= icon %>"></span>
    <span class="text">
      <span class="message"><%= message %></span><br>
      <% if (message2) { %>
      <span class="action"><%= message2 %>
      <%   if (action_js) { %>
        - <a <%= action_js %> data-event_id="<%= event_id %>" class="action <%= action_cls %>"><%= action_text %></a>
      <%   } else if (action_url) { %>
        - <a href="<%= action_url %>" data-event_id="<%= event_id %>" class="action <%= action_cls %>"><%= action_text %></a>
      <%   } %>
      </span>
      <% } else if (type === 'nagify') { %>
      <span class="action">Want to <a href="<%= action %>" data-event_id="<%= event_id %>" class="action notification-nag <%= action_cls %>">do it now?</a></span>
      <% } else if (undo) { %>
      <span class="action">A mistake? <a href="#" data-event_id="<%= event_id %>" class="action notification-undo <%= action_cls %>">undo</a></span>
      <% } %>
    </span>
    <a href="#" data-type="<%= type %>" class="notification-close"><span class="icon-x"></span></a>
  </div>
"
```

...and then uses jQuery to [replace an existing notification](https://github.com/mailpile/Mailpile/blob/b5a8526f60a1f7ea6f3aab30ac1a4bf467179492/shared-data/default-theme/html/jsapi/ui/notifications.js#L195) or [`prepend` it](https://github.com/mailpile/Mailpile/blob/b5a8526f60a1f7ea6f3aab30ac1a4bf467179492/shared-data/default-theme/html/jsapi/ui/notifications.js#L200) to the notifications area.

Due to the newlines in the above template, jQuery inserts three nodes into the DOM:

* A text node with a newline.
* The notification div.
* A text node with a newline.

Later on, when the notification finishes, it [animates the notification sliding up and removes the div with the event ID from the DOM](https://github.com/mailpile/Mailpile/blob/b5a8526f60a1f7ea6f3aab30ac1a4bf467179492/shared-data/default-theme/html/jsapi/ui/notifications.js#L31). However, this leaves two stray text nodes in the DOM.

## Suggested Fixes

* Remove the newlines from the template.
* `trim()` the template string before passing it to jQuery.

jvilk:
Fixes a memory leak.

Fixes #1931.

### DIFF
--- a/shared-data/default-theme/html/jsapi/ui/notifications.js
+++ b/shared-data/default-theme/html/jsapi/ui/notifications.js
@@ -191,13 +191,16 @@ Mailpile.notification = function(result) {
   // Show Notification
   var $elem = Mailpile.cancel_notification(result.event_id, undefined, 'keep');
   var notification_template = Mailpile.unsafe_template($('#template-notification-bubble').html());
+  // Remove excess whitespace from notification to avoid creating TextNodes in the
+  // DOM. Fixes a subtle memory leak (https://github.com/mailpile/Mailpile/issues/1931).
+  var notification_html = notification_template(result).trim();
   if ($elem) {
-      $elem.replaceWith(notification_template(result));
+      $elem.replaceWith(notification_html);
   }
   else {
       var bubbles = $('#notification-bubbles');
       if (bubbles.children().length < 15) {
-          bubbles.prepend($(notification_template(result)).slideDown('normal'));
+          bubbles.prepend($(notification_html).slideDown('normal'));
           $('.notifications-close-all span').show();
       }
   }


jvilk:
I went with the `trim()` approach since any fix I make to the template may be undone by text editors that automatically append newlines to the end of documents. `trim()` is robust to this issue.

BjarniRunar:
This looks good to me. Thank you!

PR #1971, opened 2017-11-24
Add new translation strings

JocelynDelalande:
(ran make genmessages)

Will transifex pick new messages to translate by itself ?


BjarniRunar:
I think it will. Blindly merging. ;)

JocelynDelalande:
> I think it will. Blindly merging. ;)

Thanks :-).

I see an update date of the mailpile.pot on January 2016. on that page : https://www.transifex.com/otf/mailpile/content/

I guess that means that transifex did not pick up the new strings from today merge… But I am never sure with Transifex interface :x.

Anyone with some more Tx-fu to confirm/infirm ?

Jocelyn

Issue #877, opened 2014-09-06
Need to be able to run more than one Mailpile at a time on the same machine


BjarniRunar:
Currently, Mailpile listens on port 33411 by default. This means that without manual configuration, it is not easy to launch two instances of the app at once on the same machine. This will frustrate users in at least the following cases:
- Multiple users on a VPS or home server
- Users sharing a Mac and using the "switch users" feature (also applies on Linux and Windows)
- Users who want to run more than once instance of Mailpile for their own use

Possible solutions to this include:
1. Choose the port at random on startup (ok for desktop, bad for shared servers)
2. Choose the port at random on setup (ok for desktop, tolerable for shared servers)
3. Use a local dispatcher that proxies by hostname (requires varying hostnames and a registration protocol, might be able to use PageKite, great for large deployments, barely workable on desktop)
4. Try to use the standard port, but move to another one if it is unavailable (ok for desktop, bad for servers)

Of these, 2. and 3. are probably the most appropriate, with 2. being preferred on a desktop and 3. in a multipile type environment.

All of these have the drawback that we can no longer tell the user ahead of time which port Mailpile will be listening on, which has a negative effect on the ability to document things (but a positive effect on security). This is probably unavoidable, though.

BjarniRunar:
This issue is in part addressed by mailpile-admin.py, which adds a management and proxy layer via. Apache. So this issue is at least partially solved on Linux servers (assuming folks are happy to use Apache), but still unresolved in all the desktop environments (Linux included).

BjarniRunar:
This is now "good enough", we know how this will work both on desktops and in "multipile."

gerroon:
Sorry to bump this but I d onot get how this is achieved atm. I am trying to reun multiple Mailpiles on the same pc.

kubrikseu:
_Regardless of whether Mailpile is currently able to achieve this, you can easily achieve this yourself. For this, I am assuming you run Linux or Unix._

### Creating multiple users

Create multiple users for Mailpile, as much users as you want instances. Eg:

```
# useradd -b /home -m mailpile1
```

Make sure you have `tmux` installed and available to you. Login as the first user and start Mailpile:

```
# su - mailpile1
$ tmux new -s mailpile1
$ mailpile --www=127.0.0.1:33411 --wait
```

[...]

PR #1976, opened 2017-12-14
Processlock

JackDca:
This pull request addresses #1622 (see more comments there).
Implements an interprocessor lock to stop two Mailpiles from using the same     profile directory at the same time.
Saves the PID in a file.
Checks for conflicting use of web server port.
Implement lock to allow outside processes to read mailpile.rc safely.


JackDca:
After further testing, the check for the web server port should be revised.

The port can be changed, and the new port saved to mailpile.rc, with 

./mp --www=localhost:35001

but the code in this PR checks for the availability of the previously configured port and terminates before the port number is changed. This needs to be fixed.

BjarniRunar:
Yes, I suggest dropping the port check completely and handling that in a separate change.

I am not sure we want Mailpile to bail out just because the port is unavailable, since the user can fix that by running a www command on the Mailpile CLI, in cases where the app is run interactively.

BjarniRunar:
Ideally (and this is too big for this change), what we want is for Mailpile to record to a file, in addition to its PID, the port it is listening on.

Then running a second Mailpile, could detect that the "master" is already running, prompt the user to log in to that process and provide a client interface (via REST RPC) instead of the native one. Since we already have a working REST-based CLI in the web UI, there's no reason we can't have one in the terminal as well.

BjarniRunar:
... and the reason we would record the port instead of reading it from the config, is the user might want the port to just be dynamic and change on every run. That would be best from a security point of view, and would also solve the problem when multiple users on the same machine run Mailpile at once. But again, this is beyond the scope of this change.

JackDca:
OK, but what's about the splitting of the argument list? Is this needed now?

Three more ideas:

- Allow the profile to be specified on the command line somehow - right now environment variable MAILPILE_PROFILE or MAILPILE_HOME needs to be specified which feels clumsy.

- Provide a "discover" plugin which would allow a running Mailpile to check for other profile directories and return a table of profile names and their http ports to the console and/or a JSON on the web i/f.
It would look in /home/jack/.local/share/Mailpile/*
and maybe /home/\*/.local/share/Mailpile/* (where permissions allow it) (or the equivalent on Windows)
A configuration variable would enable this always/only when logged in/never. I've already got some of the code for this.

- Provide a "dispatch" plugin which would return an http redirect to the appropriate port when the page http://localhost:33411/dispatch/[profilename] is accessed. Again, a config variable would enable this.

BjarniRunar:
We are pretty off-topic for a changelist, but okay... :)

Specifying an environment variable is no harder than a command line argument, on Unix: "VAR=foo command" instead of "command --VAR=foo". Not sure about Windows, but on Windows there will be a GUI wrapper that hides this anyway (if we expose the feature at all) so it's irrelevant. Since Mailpile's internal design doesn't make this easy to do, at all, I don't think this is worth doing.

The discovery plugin is an interesting idea, but it feels like a rabbit hole of needing to then launch the other mailpiles and interact with other processes... as it is profiles aren't very useful and I am OK with them being a hidden / secret easter-egg type feature for now. Of course, if you have working code, that changes things and I suggest you open up a new issue to discuss further in that case.

A plugin is the wrong place for a dispatcher. A dispatcher by its nature needs elevated privileges (it needs to see into multiple users' home directories) and as such should be as small and simple as possible, not part of a big blob of complexity like Mailpile proper. Also, a redirect like that simply won't work in many cases since redirects are handled by the browser and in common multi-user configurations the browser won't be able to connect directly to the port anyway. Note that the Apache integration includes a dispatcher already, and pagekite-based (or tor-hidden service based) exposure don't need a dispatcher at all.

BjarniRunar:
Regarding --login and the argument list split - this is necessary because logging in changes which commands exist and are valid, which plugins are loaded depends on the user's configuration which is unavailable until login.

The getopt processing will complain about invalid command line arguments before doing anything, so if you validate the entire list of arguments before login, you won't be able to run certain commands from the CLI.

JackDca:
Sorry for straying off-topic - will continue that discussion on the page for #877.
After further testing, this PR appears to break the Windows version (maybe related to the appdirs package???). Will delete the web server port check and (hopefully) fix the Windows problem when time permits - until then this should not be merged.

JackDca:
This push replaces the previous version with a simpler version, implemented almost entirely in manager.py, which fixes some problems with the first version.
- An interprocess lock prevents two Mailpiles from using the same profile directory at the same time. The fasteners InterProcessLock behaves differently under Windows (compared to Gnu-Linux) so a workaround is implemented to make it work under Windows.
- The PID is saved in a file.
- An interprocess lock allows safe read by an external process of the public configuration data in mailpile.rc. (Not absolutely safe for the external process because the lock is set to time out after 2 seconds so the external process can't block Mailpile.) A workaround makes this work under Windows.
- Per Bjarni's recommendation, Mailpile does not terminate if another process is already using its http port (e.g. localhost:33411). Instead it disables its http server. The user can then use the www host:port command to change to another port. (Note that the current master terminates with a messy walkback in this situation, so this is an improvement.) This also includes a workaround for Windows.

BjarniRunar:
I'm merging this locally, but making the following changes:

   * Removing the pid file creation; there's already a --pidfile= flag available for that which also lets the caller decide where to put the PID file.
   * Removed monotonic from requirements.txt and requirements-dev.txt, we are not using it directly.
   * Renamed things and cleaned up a bit for clarity (removed the need for the .have workaround)
   * Added fasteners to the requirements of our Debian package

Thanks for doing this!

JackDca:
Thanks for your input. You are teaching an old C++ programmer to be more Pythonic! Sorry I missed the --pidfile= flag.

Have you tested this on Windows? If not I will do so for completeness. I find that Windows can be full of surprises.

BjarniRunar:
That's the spirit! :grin: I should probably have mentioned the --pidfile flag earlier, but I was curious to see if your way was solving something it wasn't.

I have not tested it on Windows, no, I kept your basic technique unchanged, I think. But thank you for taking the time, it's good to know someone is testing Windows semi-regularly now. I agree, Windows can indeed be very surprising and Mailpile has some funny work-arounds as a result.

PR #1988, opened 2018-01-22
Issue 1813

kenjones21:
Moves upload progress bar to notification. No longer stalls at 100%, displays "Processing upload..." instead. When an upload fails, notification should tell user (progress bar removed), although I don't know how to test this. When upload is complete, notification tells user and is flagged complete ("c"), so it disappears after 8 seconds. 

I used file id to uniquely identify each upload notification. There may be a better way to do this.


BjarniRunar:
Hey, thanks for working on this!

I will take a closer look as soon as I can (I'm juggling a few different things so it may take me a few days).

BjarniRunar:
Sorry about the delay! I've merged this into my local tree, will push back up once I've tested a bit.

PR #1989, opened 2018-01-23
more traivs friendly way of installing packages

stephengroat:
removes need for sudo


BjarniRunar:
Looks good to me, thanks!

PR #1994, opened 2018-02-02
mailpile/util/CleanText: allow '-' in NONPATH

mrngm:
This pull requests adds `-` to the allowed characters in the `NONPATH` specification of the `CleanText` class.

### DIFF
--- a/mailpile/util.py
+++ b/mailpile/util.py
@@ -973,7 +973,7 @@ class CleanText:
                                         set(range(ord('0'), ord('9') + 1)) -
                                         set(range(ord('a'), ord('z') + 1)) -
                                         set(range(ord('A'), ord('Z') + 1)) -
-                                        set([ord('_'), ord('/')]))])
+                                        set([ord('-'), ord('_'), ord('/')]))])
 
     def __init__(self, text, banned='', replace=''):
         self.clean = str("".join([i if (((ord(i) > 31 and ord(i) < 127) or


BjarniRunar:
Could you comment on why? It's so long I wrote that, I don't remember whether I left - out deliberately, but I'm interested in your motivation here. :)

mrngm:
I was trying to access Mailpile through a reverse nginx-proxy, like this:

```
    location /example-mail/ { 
        auth_basic "Authenticate please";
        auth_basic_user_file auth.passwd;
        proxy_pass http://127.0.0.1:33411;
        proxy_set_header X-Forwarded-Host $host;
        proxy_set_header X-Forwarded-Server $host;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    }   
```
(I did not have a spare domain such that it could run on `domain.tld` itself)

Upon the first access, it gave an `Internal error` when requesting https://domain.tld/example-mail/ (as I did not yet change `sys.http_path`). Unfortunately, trying to set `/example-mail/` as `sys.http_path`, the only error message I got was `ValueError: Invalid value for config/sys/http_path: '/example-mail/'` (as it turned out, the `'` were not needed, but the error message was too vague to correctly locate the source of the problem).

After some code browsing (and editing such that I got to see the _exact_ exception), it turned out that `CleanText` filters on a list of banned characters, and the `webroot` filter allows `[0-9a-zA-Z_/]`, but not `-`. 

Since a dash is perfectly valid path element, I suggest to add it.

(on a somewhat related note, https://github.com/mailpile/Mailpile/blob/master/mailpile/config/base.py#L461 is the culprit for not seeing the exception thrown: `ValueError: Invalid web root '/example-mail/'`)

BjarniRunar:
Thank you for explaining! I'm going to refresh my memory and double-check the code's assumptions, but this sounds both reasonable and fine.

BjarniRunar:
Thank you!

Issue #1993, opened 2018-01-31
Command browse_or_launch is obsolete, should be removed


JackDca:
The idea of the browse_or_launch command appears to be to check if another Mailpile is running on the same pile; if not, then launch this Mailpile normally; if so, then start the OS's browser (to allow the user to interact with the other Mailpile) and terminate this Mailpile.

This is important, especially if as per #877 there are multiple Mailpile data sets ("piles") and/or processes on one machine and/or if port numbers are assigned randomly, since the user may not know the port number in any of those situations.

Processlock PR #1976 breaks browse_or_launch since running two Mailpiles on one pile will cause immediate termination of the second Mailpile when it detects that the pile is already locked, before reading the configuration files, so it can't start the browser.

However it appears that browse_or_launch was already broken before PR #1976. The attached file shows results in three different situations, and the browser is not started in any of them.

A decision should be made as to whether to just abandon or fix browse_or_launch.

Fixing it involves some complications since it implies that, even with the pile locked by another process, Mailpile would have to read the public configuration file mailpile.rc to determine the port on which to launch the browser.  PR #1976 does anticipate another process reading the mailpile.rc, but allowing the second Mailpile to continue executing to read it and start the browser, while guaranteeing that it does not write anything to the pile, is not straightforward.

[BrowseOrLaunchLog.txt](https://github.com/mailpile/Mailpile/files/1682276/BrowseOrLaunchLog.txt)

BjarniRunar:
I think this API command is obsolete and should just be removed. The desktop integration work has taken a path which doesn't use it. Renaming issue to match.

noahhaasis:
Can I do this commit?

BjarniRunar:
Go for it!

Jagroop-Singh:
Is this open or closed? It sounds like noahhaasis has already solved the issue and closed it, and somehow it reopened?

Jagroop-Singh:
@BjarniRunar

PR #1995, opened 2018-02-02
Tools for managing multiple Mailpiles on one machine re #877

JackDca:
This PR attempts to address some of the issues identified in #877.

- It adds random port selection to the www command, invoked by specifying port number 0. New public parameters http_port_min and http_port_max put limits on the new port number. The discover command code (if available - see below) is used to check against a list of port numbers already in use by other piles (where permissions allow) and the port is also checked to see that no process is actively listening on it.
- It provides a "discover" command. This uses the full path to the data directory (workdir) of the present Mailpile as a template, and checks for other piles belonging to the present user and other users (where permissions allow). Parameters are provided to limit the search to specific user(s) or pile names(s). The result is a table of configuration parameters extracted from mailpile.rc files.
- The term "profile" was being used in two different ways in manager.py: to mean all the configuration parameters associated with one receiving email address; and, to mean all user data accessible to one Mailpile instance (as in environment variable MAILPILE_PROFILE). The first meaning was used extensively in the UI and in the configuration file, so in the code the term "pile" was used to replace the term "profile" where the second meaning was intended. Variables called "pile" are used for the base name (last component) of the data directory path. Consistent with this, a new environment variable MAILPILE_PILENAME is checked and takes priority over MAILPILE_PROFILE.

The www random port selection can be used to create a new pile on a random port or to access an existing pile on a random port on every launch (#877 option 1)

	export MAILPILE_PILENAME=alicemail ; ./mp --www localhost:0 --interact

Once a pile is set up, it can also be accessed without changing the port number from the one used at setup (#877 option 2):

	export MAILPILE_PILENAME=alicemail ; ./mp

The user can also get a list of piles on the machine. At present Mailpile's default permissions only allow access to piles belonging to the same user but the command will look for everything that is accessible:

	./mp discover

or with optional parameters it can be limited to specific users and/or pile names:

	./mp discover --user=alice,bob --pile=default,alicemail

Comments:

- It would be good to allow launching of the web browser to access an already running Mailpile. When Mailpile is not at the default port location the user may not know the port. Unfortunately browse_or_launch is broken at present (#1993) so this doesn't work:

	export MAILPILE_PILENAME=alicemail ; ./mp browse_or_launch

- Comments would be appreciated on the format of the argument list for discover. There seem to be several possibilites (leading -- on command? -- on options? use = or two seperate strings?).

- Should authentication be required before discover can be used?

All the above is presented for discussion. Any and all comments are welcome

### DIFF
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -187,9 +187,11 @@ def Main(args):
         session.main = True
         try:
             CatchUnixSignals(session)
-            config.clean_tempfile_dir()
+            config.clean_tempfile_dir()     # *** What if this is 2nd Mailpile.
             config.load(session)
+            print 'TTTTT'
         except IOError:
+            print 'UUUUU', config.sys.lockdown
             if config.sys.debug:
                 session.ui.error(_('Failed to decrypt configuration, '
                                    'please log in!'))
@@ -199,7 +201,7 @@ def Main(args):
         session.ui.error('Access denied: %s\n' % e)
         sys.exit(1)
 
-    print 'AAAAA' # *** DEBUG
+    print 'AAAAA', config.sys.lockdown
 
     try:
         try:
@@ -246,6 +248,11 @@ def Main(args):
         traceback.print_exc()
 
     finally:
+    
+        print 'BBBBB',config.sys.lockdown
+        if isinstance(config.lock_workdir, str):
+            pass
+    
         if readline is not None:
             readline.write_history_file(session.config.history_file())
 
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -187,11 +187,9 @@ def Main(args):
         session.main = True
         try:
             CatchUnixSignals(session)
-            config.clean_tempfile_dir()     # *** What if this is 2nd Mailpile.
+            config.clean_tempfile_dir()
             config.load(session)
-            print 'TTTTT'
         except IOError:
-            print 'UUUUU', config.sys.lockdown
             if config.sys.debug:
                 session.ui.error(_('Failed to decrypt configuration, '
                                    'please log in!'))
@@ -200,8 +198,6 @@ def Main(args):
     except AccessError, e:
         session.ui.error('Access denied: %s\n' % e)
         sys.exit(1)
-
-    print 'AAAAA', config.sys.lockdown
 
     try:
         try:
@@ -248,11 +244,6 @@ def Main(args):
         traceback.print_exc()
 
     finally:
-    
-        print 'BBBBB',config.sys.lockdown
-        if isinstance(config.lock_workdir, str):
-            pass
-    
         if readline is not None:
             readline.write_history_file(session.config.history_file())
 
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -199,6 +199,8 @@ def Main(args):
         session.ui.error('Access denied: %s\n' % e)
         sys.exit(1)
 
+    print 'AAAAA' # *** DEBUG
+
     try:
         try:
             if '--login' in args:
--- a/mailpile/config/defaults.py
+++ b/mailpile/config/defaults.py
@@ -35,6 +35,8 @@ CONFIG_RULES = {
         'history_length': (_('History length (lines, <0=no save)'), int,  100),
         'http_host':     p(_('Listening host for web UI'),
                            'hostname', 'localhost'),
+        'http_port_min': p(_('Min random port for web UI'), int,        32769),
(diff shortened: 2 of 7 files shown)


JackDca:
An IRC discussion between @BjarniRunar, @karlp and @JackDca resulted in these comments. These were not unanimous and I have added my own interpretation in places - please correct if I have misrepresented the meaning!

1. Better to separate this kind of tool from the main part of Mailpile.
2. A possible model is shared-data/multipile/mailpile-admin.py which does for Multipile what the new tool would would do for multiple users on one machine (each with their own single-user Mailpile).
3. One use case:  family with a home computer and lots of remote devices, or maybe a small office; basically an setting where there's no sysadmin and the OS might be Windows (and don't forget Apple).
4. We want to actively discourage the creation of a 'nix OS userID (e.g. "mailpileadmin") that hosts multiple users' piles in its home directory.
5. At present the only supported implementation providing remote access is mailpile-apache2.
6. BjarniRunar is developing a remote access capability using native Python TLS with Pagekite to make it remotely accessible. The details need to be defined before the requirements of a management tool can be stated.
7. Windows may be different in that some Windows machines are used by multiple users on the same OS userID.
8. Permissions are an issue. Before designing the tool, need to define what permission setup would work and what is the default permission configuration on various OS's. Windows permission structure is a mystery to the participants in this discussion, except that the firat user configured after a new install has some admin powers.
9. There is some discomfort with one Mailpile snooping in other user's home directories.
10. Possible the "master" Mailpile could use a fixed port so the ports that give access to other piles can be looked up there.

Some additional comments (mine):
Re: 2. This tool could be integrated with a systray applet (like the one already implemented for Windows) so that in the case of a forgotten port number, access through the desktop would be possible.
Re: 4. This would definitely cause problems since the GnuPG data would be shared between users!
Re: 6. The code in this PR lists the sys.proxy.port and sys.proxy.protocol for each pile though (as BjarniRunar pointed out) this is probably not useful. The mailpile.rc options that are listed by discover are table-driven, so this was intended as an example of how remote access ports might be discovered. That is, the remote access URL could be stored in mailpile.rc options so that discover could list it.
Re: 10. "Fixed" could mean "fixed for this machine", it might not be 33411. Using high port numbers instead of ICANN sanctioned numbers is "security by obfuscation" and cannot be relied upon.  BUT, an open fixed port (especially 22) attracts brute force attacks which are annoying even if they never succeed; it also facilitates mass surveillance by traffic analysis. The idea is to have each user know that, if they forget/erase/change their own url/port number, they can go to a fixed port number and look it up with discover.

As always, comments solicited!

BjarniRunar:
I want to apologise for not having responded to this. I think it has proven too big a bite for me to chew, given other demands for my attention and time.

I like some of this, but some of this I am less keen on - and since this is a markedly different approach from what we're actually doing in the desktop integration sprint, I've been unsure which parts to keep and which parts to reject.

My current plan is to merge this locally and then revert/undo the bits I don't want to keep.

I think in the future, for such large changes it would be really helpful if you could break them up into smaller pieces - the bullet points you gave in your original description were pretty obvious ways to break this up, and I probably would have merged 2/3 and them discussed the third if they'd been presented that way.

JackDca:
OK - glad that it is of some use.

JackDca:
@BjarniRunar - re your last comment - you did not state which 1/3 of the changes you did not like, but from you other comments it appears it is the integration of the discover command with the main part of Mailpile. Agreed it's out of place there. Based on that, here is a suggested resolution of this PR:
1. Retain the random port selection capability.
2. Retain the change in variable naming from "profile" to "pile" to help readability for future maintainers. This is transparent to the user **except** it adds the environment variable MAILPILE_PILENAME (is there a more elegant name?) which takes precedence over MAILPILE_PROFILE if present.
3. Move the discover functionality from the Mailpile CLI to a stand-alone scripts/mailpile-discover.py (consistent with the naming of mailpile-decrypt.py) so it is available as a tool for developers/sysadmins.

If this is done there is a decision necessary re the validation of randomly selected ports (1.). The port will be validated by as in the present PR by checking that no executing process is listening on the same port. Should it also be checked, as in the present PR, with the discover code to see if another pile, not currently executing, has selected it? It seems logical to do everything possible to avoid port conflict. Options:
- Extract the discover_piles() method from the discover command and put it in mailpile/config/discover.py where it can be referenced by the port number validation code in mailpile.plugins.core.RunWWW(), and by the scripts/mailpile-discover.py .
- Don't attempt to check other mailpile configurations for port conflict.

PR #2004, opened 2018-02-14
Fix display of recipients image in search listing

JocelynDelalande:
See commit messages

That follows 8fc7deb


BjarniRunar:
Thank you, but I disagree with this!

I deliberately left in the avatar of the sender, as that tells you which account was used to communicate. which is very important information.

I just wish the reply arrows pointed the other way, so it would be more visually clear that the message was going from the "avatar" to the recipient. It works perfectly for forwards and initial contact...

If you feel I'm badly wrong here, maybe let's discuss further on #2005 (renaming that issue)?

PR #2012, opened 2018-02-20
Fix memory leak in autoajax caused by extraneous whitespace

jvilk:
This leak is identical in nature to #1931, and was automatically found by [BLeak](https://github.com/plasma-umass/bleak) while running it on Mailpile again.


BjarniRunar:
Awesome, thanks!

PR #2018, opened 2018-03-08
Fix interprocess lock that does not work on Windows

JackDca:
This is a one line update of PR #1976. Its final version tests for a lock on the workdir like this:

`if isinstance(self.lock_workdir, str):`

It appears that under Windows, `os.path.join()` returns a unicode (not str) object when initializing self.lock_workdir, so the above test always fails.

This fix has been tested on Windows7.

### DIFF
--- a/mailpile/config/manager.py
+++ b/mailpile/config/manager.py
@@ -217,7 +217,7 @@ class ConfigManager(ConfigDict):
             os.makedirs(self.workdir, mode=0700)
 
         # Once acquired, lock_workdir is only released by process termination.
-        if isinstance(self.lock_workdir, str):
+        if not isinstance(self.lock_workdir, fasteners.InterProcessLock):
             ipl = fasteners.InterProcessLock(self.lock_workdir)
             if ipl.acquire(blocking=False):
                  self.lock_workdir = ipl
--- a/packages/windows/repackage-linux.sh
+++ b/packages/windows/repackage-linux.sh
@@ -46,6 +46,11 @@ GNUPG_FILE_ALL="gnupg-w32-2.2.1_*.tar.xz"
 
 # End of configuration strings.
 
+# Capture path to package script directory, go there.
+set -e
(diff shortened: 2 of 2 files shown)


BjarniRunar:
You've also got a package version change in there, which doesn't seem to belong.

Incidentally, regarding that, the package version you're configuring there does not match the numbering used for our Debian packages - I have deliberately been calling the 1.0 rc series "0.99.x" so it will be obvious that 1.0.0 is newer. Some ways of sorting and comparing versions would tread 1.0.0rc1 as newer than just 1.0.0.

JackDca:
Hi Bjarni - OK, I found your comment on the website about the version numbers. (Always something to learn!) By that system I would have called it 0.99.2+, the + indicating a minor unofficial update to 0.99.2 and not a full fledged release. 

Maybe I can automate the versioning to produce something like the mailpile_0.99.2~dev201803080351-1_all file name that your official nighties use. In fact that is a necessary update to the packaging script since right now the version is entered manually. How is the nightly version name generated?

Some versioning systems use the tilde ~ character because it is sorted to _precede_ every other character. That is 1.0.0~rc2.xxx precedes 1.0.0.xxx. I'm not clear how that works since ~ is 7Eh so you'd expect it to come _after_ everything else in the collating sequence.

BjarniRunar:
Check out `scripts/version.py` :grin: 

I don't know if this is a record! It should be pretty close! I am doing a terrible job tracking critical metrics like this. :cry:

JackDca:
Rather than reverting the version change, this update gets an appropriate version using `scripts/version.py` so e.g. building an installer from this branch results in this file name:

Mailpile-0.99.2~processlockfix201803091151-Installer.exe

Also automated the calculation of SHA256 checksums.

(At some point we need to add signing of the installer using a certificate that Windows will recognize.)

BjarniRunar:
Awesome!

Issue #2021, opened 2018-03-26
Update Dockerfile to use our Debian repo


BjarniRunar:
The current Dockerfile is doing things "its own way."

I don't user Docker and I'm usually alone on the core team - so this virtually guarantees that the Dockerfile becomes obsolete over time. And no, I'm busy, expecting me to just learn Docker and add this to my workload already isn't going to happen. :grin: 

It would be awesome if someone could rework the Dockerfile to use a minimal Debian image, add our Debian repository and install from packages that are actually maintained.

The package should by default track the release repo, but an option to use nightly would make sense. If Dockerfiles have options, otherwise just another Dockerfile. See: https://www.mailpile.is/download/linux.html

BjarniRunar:
Thanks @rgarrigue for working on this!

First, I agree with your comments from the PR that with this pattern, an official Docker does indeed start to make sense. Since this is (aside from the sandboxing and special user account) following roughly the same patterns as our non-Dockerified linux setups, it shouldn't increase our support burden or complicate things too much.

Regarding that, I do have a couple of questions:

   * How do Docker installs like this get updated? What procedure should we recommend?
   * I notice you create VOLUMEs for the data directories, can you explain to a Docker newbie what that means? I assume this is part of not losing all your data if/when you rebuilt the container?

BjarniRunar:
And also:

    * Can the Dockerfile be parametrized to pull in the nightly repo somehow?

rgarrigue:
I guess some will help. Shortly for now

- Update is docker stop, docker pull (updates image), docker start. The app (mailpile) gotta be able to detect old version data and update it if required.

- Mounting volume is persisting data. Otherwise a container is just a process, all in memory. Kill or stop it and data is lost. The relevant folders to be mounted are usually declared with VOLUME instruction in Dockerfile.

- hub.docker.com can make a tag by git branch, and will trigger a build on a github push via webhook. Fixing a version in Dockerfile might not be necessary. But you can do it.

BjarniRunar:
Thank you! Mailpile's normal update strategy is like that of any other Debian app - updates from the repo. Is that "incompatible" with The Docker Way?

rgarrigue:
Well, to paraphrase you, Docker app's normal update strategy is is like that of any other app - updates from the repo ;-)

Don't think about docker like you'ld think about a VM. The Docker Way is

- One app by container, fix version with tag
- Update by pulling a new docker image, built around the new app version.

[...]

PR #2022, opened 2018-03-26
Reworked Dockerfile

rgarrigue:
Hi

Following #2021, here's a reworked version if the Dockerfile using official debian repo over debian buster.

Here's an automated build for this container : https://hub.docker.com/r/rgarrigue/mailpile/builds/, so you can test it. As for myself I'm running it on my own little cloud, everything seems to be working fine.

Also I'ld suggest setting up an automated official build for mailpile, it's dead simple : create a mailpile account on hub.docker.com, link to a github account (mailpile org via a token?), create automated build, fill description and done with it.

Regards,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,30 @@
-FROM alpine:3.5
+FROM debian:buster-slim
 
-WORKDIR /Mailpile
+ENV GID 33411
+ENV UID 33411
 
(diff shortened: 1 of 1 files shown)


BjarniRunar:
This looks very cool, thank you!  I have some questions, but I will move the discussion to the issue.

Issue #1566, opened 2016-02-03
Avoid garbage seach results on unclean shutdown


BjarniRunar:
Currently, if Mailpile shuts down uncleanly, this is almost guaranteed to result in garbage search results. Sequence of events:
1. Message is discovered, assigned an MID and added to search index.
2. Search index is incrementally updated, some keyword-to-MID mappings get written out.
3. Unclean shutdown, MID and metadata is not written to disk.
4. Mailpile restarts, assigns MID to a different message.
5. Search results for MID are combined from multiple mails.

In particular, step 4 is almost guaranteed to do the wrong thing if any new mail arrives while Mailpile is shut down, because messages are processed in reverse order.

Proposed fix:

On startup, scan the global posting list for MIDs that are not present in the metadata index. Assign tombstones to all the MIDs found this way, and mark deleted.

BjarniRunar:
Proposal on how to implement this:

1. Add method to postinglist.GlobalPostingList which finds the largest known MID
2. In search.Index.load(), use that method after loading the global posting list to sanity check the metadata
3. Create tombstone entries for all MIDs necessary to bring Metadata up to sync with the GPL.

JackDca:
While testing PR #2023 I noticed that mailpile.idx is often not saved to the file system until Mailpile is shut down. This seems relevant to the issue of unclean shutdowns.

Examples:

Run Mailpile, with no pre-existing pile (workdir), and create an account referencing an existing IMAP server account that contains about 200 emails. Mailpile downloads the emails, indexes them, and writes kw-journal.dat (the global posting list) within a few minutes. It does not write mailpile.idx (the metadata index) until the CLI quit command is given. This is true even if Mailpile is left running for e.g. three days and receives one or two new emails from the IMAP server in that time.

Even after a quit and restart, with new messages being received after the second start of Mailpile, an updated mailpile.idx index is not written until another quit command is entered. In that case the updated mailpile.idx is identical to the first mailpile.idx but with the metadata for the new contacts and messages appended to it.

Is this is the intended behaviour?

Data integrity in the case of unclean shutdowns might be improved if mailpile.idx were written more frequently. Are there complications related to the need to append to an encrypted file?

BjarniRunar:
Rewriting the full metadata.idx is a rare event because it is an expensive operation. Most of the time we just append changes, but appending should be relatively frequent - not just on quit. If it's ONLY happening on quit we have a bug.

JackDca:
[...]

PR #2023, opened 2018-03-28
Fix for issue #1566 - garbarge search results on unclean shutdown

JackDca:
This fix follows roughly the approach proposed in #1566.

- Create special key '1' in GPL to hold max MSG_MID in GPL.
- Update value for key '1' whenever GPL is updated.
- Check metadata index against GPL '1' value at startup,
    add ghost entries for missing MSG_MIDs.
- Block migration of GPL '1' to PostingListContainer.

The maximum MSG_MID is maintained under a "special key" '1' in the GPL. This avoids the possibility that all copies of the maximum MSG_MID might be migrated to PostingListContainers which would make the result of scanning the GPL incorrect. (This may never happen in practice but it was not obvious from studying the code that it was impossible.) Also the "special key" approach requires less number crunching. Comparing those variable length base36 values is annoying.

The code has been tested by killing Mailpile during the download of about 200 emails, then restarting. Searches were done using the web UI, and mailpile.idx was decrypted and examined. It appears to work. There's no special code needed to migrate from earlier versions. The first time that a new email is indexed with this code, key '1' will be created and immediately saved (appended) to kw-journal.dat.

However, the deleted metadata index entries still appear with the subject "(missing message)" unless they are filtered out, e.g. by adding the search term "in:inbox". Perhaps "ghost" entries should be filtered out of all search results.

### DIFF
--- a/mailpile/postinglist.py
+++ b/mailpile/postinglist.py
@@ -646,7 +646,48 @@ class GlobalPostingList(OldPostingList):
                 GLOBAL_GPL[sig] = set()
             for mail_id in mail_ids:
                 GLOBAL_GPL[sig].add(mail_id)
-
+                
+    @classmethod
+    def GetMaxMsgIdxPos(cls):
+    
+        global GLOBAL_GPL
+        with GLOBAL_GPL_LOCK:
+            try:
+                # Maximum MSG_MID in keyword index held under special key '1'.
+                max_msg_idx_pos = max([int(id, 36) for id in GLOBAL_GPL['1']])
+            except:
+                max_msg_idx_pos = None
+        return max_msg_idx_pos
+    
+    @classmethod
+    def UpdateMaxMsgMid(cls, session, msg_mids):
+    
+        if len(msg_mids) == 0:
+            return
+        try:
+            msg_idx_pos = max([int(id, 36) for id in msg_mids])
+        except:
+            return
+        
+        global GLOBAL_GPL   
+        with GLOBAL_GPL_LOCK:
+            try:
+                # Maximum msg_mid in keyword index held under special key '1'.
+                max_msg_idx_pos = max([int(id, 36) for id in GLOBAL_GPL['1']])
+            except:
+                max_msg_idx_pos = None
+            
+            if not max_msg_idx_pos or msg_idx_pos > max_msg_idx_pos:
+                msg_mid = b36(msg_idx_pos)
+                                            
+                super(GlobalPostingList, cls)._Append(session,
+                                    '', [msg_mid], sig='1', compact=False)                     
+                if GLOBAL_GPL is None:
+                    GLOBAL_GPL = {}
+                GLOBAL_GPL['1'] = set([msg_mid])
+            
+        return
+                
     def __init__(self, *args, **kwargs):
         with GLOBAL_GPL_LOCK:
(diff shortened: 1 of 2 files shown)


BjarniRunar:
Wow, cool! I love that you're diving into the guts of the app like this.

I'll take time to review this as soon as I can.

JackDca:
Glad to help! No rush on the review. Sorry about the conflict - I try to avoid that. Maybe I can fix it next week when I'm back from the holiday weekend.

BjarniRunar:
Update: I'm going to pull this down and merge in my local tree. I'm tempted to change "1" into something a bit more meaningful, but aside from that I think this looks great. I just need to play with it a bit on my machine before I push it back up.

JackDca:
I used '1' just because it makes the check in _migrate() very simple:   
len(sig) <= 1
and it also occurred to me that there might be some future need for more "special" keys which could be '2', '3', ... .   
As long as the _migrate() check is adjusted accordingly, any key is OK.

BjarniRunar:
This got merged, but GitHub didn't understand the merge. Weird.

JackDca:
My original PR had a conflict, although I fixed it. Probably this is related.

JackDca:
Hello Bjarni,

Since you have done the lion's share of the work on Mailpile it is certainly your prerogative to make the call. But allow me one final comment on the issue after which I will forever hold my peace.

My original PR had a vulnerability in that, if an unclean shutdown occurred during an append to kw-journal.dat, the updated  `max_msg_mid` might not be saved in the file. Likewise, an unclean shutdown during a full save of kw-journal.dat might result in the `max_msg_mid` not being written since its position in the file was not explicitly defined. This vulnerability is avoided by writing  `max_msg_mid` as the first line of an append (rather than the last as in my code), or as the first line of the complete file save.

There is no reason to believe that a wholesale scan of kw-journal.dat is less prone to errors or bugs than saving `max_msg_mid` in a dedicated key. The full scan is guaranteed to give the right answer **if** the file itself is free of errors and it is read correctly. But, the number of lines of code and machine instructions that impact the full contents read from kw-journal.dat is, if anything, greater than the number of lines that impact the saved `max_msg_mid`. And, you still need the dedicated `max_msg_mid` key to cover a situation where all the normal lines containing it have migrated to the PostingListContainer files, so an erroneous too-large value in the saved `max_msg_mid` will have just the same result as if you didn't do the wholesale scan.

Re the writing of mailpile.idx, I have been logging the behaviour in my production instance and I believe that there really is a concern. I'll open a new issue and you can be the judge.

Thank you for crediting my work so prominently (blush).

Good luck fighting off the flu virus!

BjarniRunar:
All I can say, is my gut feeling about what sort of things are likely to go wrong tells me that a full scan is likely to lead to a more resilient system.

Not all bugs are equally likely - the kind of bugs that make a full scan fail or come to the wrong conclusion are also likely to corrupt Mailpile's operations beyond recovery anyway. If the code is correct (and it's trivial code which I do believe is correct having tested it carefully), then we're talking about CPU and memory bugs here. Those kill us no matter what.

However, failing to fully write things to disk or the programmer simply forgetting to update the max message tracker in some new code path, is exactly the kind of bugs that we have frequently had in the past and I expect us to have again in the future. The full scan mitigates or eliminates the impact of such bugs.

Issue #1874, opened 2017-07-18
Create a Hint about the side-bar Organize mode


BjarniRunar:
The user can customize their sidebar by clicking the Organize button at the bottom of the sidebar. The organizing happens mostly by dragging tags up and down, so they end up in different places.

The fact that you can drag/drop things and move them around is non-obvious and needs explaining. The current mechanism we have for such explaining is the Hints plugin, in `shared-data/contrib/hints/` - adding some explanations there would be a good first step.

Longer term, it might make sense for the "Organize" mode to actually gray out the search result area and replace it with helpful text or images. That's a bigger task, obviously.

(thanks to @karlp for the feedback and discussion)

BjarniRunar:
Also, this would be NICE TO HAVE for 1.0, but isn't a blocker. Just sayin'. :-)

Coronon:
Should be closed

BjarniRunar:
@Coronon Correct, thank you very much for the reminder.

PR #2024, opened 2018-04-03
Sidebar Organization Hint

MikeynJerry:
Hi, I'm a beginner and student who needs to make a contribution to an open source project for my class so I've picked this project to work on. I've taken a look at issue #1874 and attempted to implement a hint that tries to explain how organizing the sidebar works.
Here's how it looks on my client:
image

I've set the timer to 6 days because it seemed like you guys were going for a hint a day and the 6th day was the next one available.

I'm sure you're all very busy, but it would mean the world to me if you could review it sometime soon! I'll gladly fix whatever is wrong with it and go back to the drawing board if necessary. (If my contribution gets accepted I get extra credit so I'm eager to help!)

### DIFF
--- a/shared-data/contrib/hints/hints.py
+++ b/shared-data/contrib/hints/hints.py
@@ -68,10 +68,8 @@ class hintsCommand(Command):
             _('Learn how to get the most out of Mailpile\'s spam filter'),
             '/page/hints/spam.html'),
 
-	# Show the user how to organize their sidebar after 6 days
-	# Remind them about it every 90 days in case they forgot how
-	# FIXME: Allow the user to say, "I know, shut up forever please."
-	('organize-sidebar', 6, 90,
+	# Show the user how to organize their sidebar after 6 days.
+	('organize-sidebar', 6, 99999,
 	    _('Rearrange your sidebar to organize how you see your e-mail'),
 	    '/page/hints/organize-sidebar.html'),
 
--- a/shared-data/contrib/hints/hints.py
+++ b/shared-data/contrib/hints/hints.py
@@ -68,6 +68,13 @@ class hintsCommand(Command):
             _('Learn how to get the most out of Mailpile\'s spam filter'),
             '/page/hints/spam.html'),
 
+	# Show the user how to organize their sidebar after 6 days
+	# Remind them about it every 90 days in case they forgot how
+	# FIXME: Allow the user to say, "I know, shut up forever please."
+	('organize-sidebar', 6, 90,
+	    _('Rearrange your sidebar to organize how you see your e-mail'),
+	    '/page/hints/organize-sidebar.html'),
+
         # Introduce Gravatar integration after 10 days, and yearly repetition.
         # Remind of the privacy implications
         ('gravatar', 10, 365,
--- a/shared-data/contrib/hints/hints/organize-sidebar.html
+++ b/shared-data/contrib/hints/hints/organize-sidebar.html
@@ -21,7 +21,7 @@
         <li>{{_("Bills")}}</li>
         <li>{{_("Groceries")}}</li>
         <li>{{_("Personal")}}</li>
-        <li>{{_("Work")}}</li>
+        <li>{{_("Work")}}</li> 
       </ul>      
     </ul>
   </ul>
--- a/shared-data/contrib/hints/hints/organize-sidebar.html
+++ b/shared-data/contrib/hints/hints/organize-sidebar.html
@@ -1,0 +1,31 @@
+{% extends "layouts/" + render_mode + "-tall.html" %}
+{% block title %}{{_("Organizing Your Sidebar")}}{% endblock %}
+{% block content %}
+<div class="content-normal" style="max-width: 46em;">
+  <h1><span class="icon-tag"></span>
+    {{_("Organizing Your Sidebar")}}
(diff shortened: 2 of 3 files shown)


shared-data/contrib/hints/hints/organize-sidebar.html
```diff
@@ -0,0 +1,31 @@
...
+    <li>{{_("To customize each tag individually, click on one of the gears next to any tag.")}}</li>
+    <ul style="margin-left: 2em; list-style: circle">
+      <li>{{_("Under Tag Settings you can change the name and color of a tag as well as where it's located and if you want it in your searches.")}}</li>
+      <li>{{_("Under Technical Settings you can make sub-tags so that they're more organized. For example:")}}</li>
+      <li>{{_("Receipts")}}</li>
+      <ul style="margin-left: 2em; list-style: circle">
+        <li>{{_("Bills")}}</li>
+        <li>{{_("Groceries")}}</li>
+        <li>{{_("Personal")}}</li>
+        <li>{{_("Work")}}</li>
```
BjarniRunar: You're missing an end-of-list `</ul>` here

shared-data/contrib/hints/hints/organize-sidebar.html
```diff
@@ -0,0 +1,31 @@
...
+<div class="content-normal" style="max-width: 46em;">
+  <h1><span class="icon-tag"></span>
+    {{_("Organizing Your Sidebar")}}
+  </h1>
+  <p>
+    {{_("Mailpile lets you organize your mailbox the way you want it be.")}}<br>
+    {{_("You can start customizing your experince by rearranging the sidebar on the left.")}}
+  </p>
+  <ul class="list" style="margin-left: 2em; list-style: disc">
+    <li>{{_("In the bottom left, click ")}}<b>{{_("Organize.")}}</b></li>
```
BjarniRunar: We try to avoid breaking sentences into fragments, since different languages prefer a different order. In cases like this though, because of the formatting, it can be hard to avoid. Generally what we've done is follow this pattern:

In the bottom left, click: Organize.

In the code that would look like this:

    {{_("In the bottom left, click:")}} <b>{{_("Organize")}}.</b>

Note the differences in punctuation and spacing - I put the : inside the string, as a hint to the translator that something will come after, but took the white space out (since translators might accidentally remove it). However, I've removed the period from the translatable word "Organize", as that same word may be used/translated elsewhere without the punctuation being appropriate.

shared-data/contrib/hints/hints/organize-sidebar.html
```diff
@@ -0,0 +1,31 @@
...
+    <li>{{_("From here you can click and drag any of the tags (Inbox, Outbox, etc.) up or down to change its order in the sidebar.")}}</li>
+    <li>{{_("To customize each tag individually, click on one of the gears next to any tag.")}}</li>
+    <ul style="margin-left: 2em; list-style: circle">
+      <li>{{_("Under Tag Settings you can change the name and color of a tag as well as where it's located and if you want it in your searches.")}}</li>
+      <li>{{_("Under Technical Settings you can make sub-tags so that they're more organized. For example:")}}</li>
```
BjarniRunar: Please break this into two translatable sentences; make "For example:" stand on its own.

shared-data/contrib/hints/manifest.json
```diff
@@ -15,7 +15,8 @@
         "/page/hints/deletion.html": {"file": "hints/deletion.html"},
         "/page/hints/backups.html": {"file": "hints/backups.html"},
         "/page/hints/gravatar.html": {"file": "hints/gravatar.html"},
-        "/page/hints/spam.html": {"file": "hints/spam.html"},
+        "/page/hints/organize-sidebar.html": {"file": "hints/organize-sidebar.html"},
+	"/page/hints/spam.html": {"file": "hints/spam.html"},
```
BjarniRunar: Did the indentation change here? Or are you using TAB characters instead of spaces?

shared-data/contrib/hints/hints.py
```diff
@@ -68,6 +68,13 @@ class hintsCommand(Command):
             _('Learn how to get the most out of Mailpile\'s spam filter'),
             '/page/hints/spam.html'),
 
+	# Show the user how to organize their sidebar after 6 days
+	# Remind them about it every 90 days in case they forgot how
+	# FIXME: Allow the user to say, "I know, shut up forever please."
+	('organize-sidebar', 6, 90,
```
BjarniRunar: Let's just never remind the user about this one, I don't think it's necessary. Setting the interval to 99999 should do the trick. :-)

MikeynJerry:
Okay, I committed again with the changes you've requested. For some reason it's not saying the </ul> one is outdated but I think that's because it ended up on line 25 and you only referenced 21-24 but I did add a </ul> tag! Please let me know if this doesn't fix the issues you found.

Also, I think an argument could be made for reminding the user about organizing the sidebar at some point because if someone decides to not organize their sidebar now, they might want/need to later and not remember how. That might be an edge case though so for now, I think 99999 is fine.

MikeynJerry:
Okay, everything should be fixed now. Please let me know if it's not!

BjarniRunar:
Thanks for doing this!

Issue #2027, opened 2018-04-04
recipients full email not displayed


hiteshramani:
Hi,

When I type in the recipients email id, only the user name is displayed and the domain name. Is this done on purpose? I feel the entire email id should be displayed for user's convenience, also, this might help to avoid any ambiguity that may occur due to a recipient user 2 different domains with the same username.

BjarniRunar:
I don't understand this issue. "Username and the domain name" - that's the entire e-mail address, what more are you hoping for?

If you are talking about how it abbreviates it in the To: field AFTER you have chosen the recipient, yes, I see how that can be confusing, but since you just typed it in yourself and you can hover to double-check it's still what you want I am not sure there is much of an issue here. Is that what you are talking about?

hiteshramani:
@BjarniRunar Yes I meant how it abbreviates it in the "To:" field post a new/old recipient has been chosen, typed. If you see Gmail/Outlook they display the entire email address:

Gmail
![gmail_screen](https://user-images.githubusercontent.com/6819461/38312492-8d1ef444-383f-11e8-998e-74eccd448a87.png)

Outlook
![outlook_screen](https://user-images.githubusercontent.com/6819461/38312515-9b3713cc-383f-11e8-8c51-118eab8ca70c.png)

Mailpile
![mailphile_screen](https://user-images.githubusercontent.com/6819461/38312559-b1bbd722-383f-11e8-9160-bb5b69117a29.png)

Of course, this is not a critical requirement but it does help in the experience. I raised the issue with the point of view of a beginner contributor, so if you feel this might be a good add-on, I'd like to take it up and implement.

BjarniRunar:
If you would like to try and modify this, I would happily merge that pull request. I agree with you that showing the full e-mail makes more sense.

PR #2043, opened 2018-04-11
Recipients email ID

hiteshramani:
This PR attempts to resolve the issue #2027 

- This intends towards a better user experience wherein when an email is being composed, if the recipient is already a contact, the name of the contact is displayed, but if the recipient's email id isn't a contact for the account of the user, the entire email id will be displayed.

- Also, while digging deep and checking for other places wherein this might require an edit, I found some unclosed tags in search_item.html.


BjarniRunar:
This contact change looks good, thanks!

However, your closing-opened-td-tags change is incorrect. The final `</td>` is after the `{% endif %}`, there are different opening tags inside the different `{% if ... %}` blocks because different classes are being set. If you could remove/revert that from this PR (or submit another one without it), I'll be happy to merge.

hiteshramani:
Okay, I'll revert those changes in this PR itself.

BjarniRunar:
Excellent! Thank you!

PR #2051, opened 2018-04-18
Add a setting to configure AJAX request timeout

### LINKED ISSUE #2050: Timeout opening large emails
I get system status reports from a few servers over email, and the messages can get quite long. If I have a particularly long email (maybe listing a few thousand changed files), mailpile will be chew on it for a while with a "Working..." message in the notifications area followed by:

> Mailpile timed out...
> Check your network?

And then

> Mailpile is unreachable.
> Check your network?

During this time the python process on my machine consumes 100% CPU. Is there a setting to tweak the timeout value somewhere? I realize that this kind of email is definitely not a standard use case, but I thought I would bring it to your attention anyway.

Thanks again for the great software. :)

dominic-p:
Here is my initial stab at a configurable AJAX request timeout. It seems to be working pretty well for me in testing. Let me know if you see any issues.

The one change that could create issues is adding a timeout to the autoajax.js request. I'm not sure what the default timeout is when none is specified. From a quick search, it seems like the default is no timeout at all. So, adding one maybe cause requests that used to work before to fail now.

Fixes #2050.

### DIFF
--- a/mailpile/config/defaults.py
+++ b/mailpile/config/defaults.py
@@ -39,6 +39,7 @@ CONFIG_RULES = {
         'http_port':     p(_('Listening port for web UI'), int,         33411),
         'http_path':     p(_('HTTP path of web UI'), 'webroot',            ''),
         'http_no_auth':  X(_('Disable HTTP authentication'),      bool, False),
+        'ajax_timeout':  p(_('AJAX Request timeout'), int,              10000),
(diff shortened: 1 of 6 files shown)


BjarniRunar:
Sorry about the delay! I've merged this into my local tree, will push back up once I've tested a bit.

dominic-p:
Thanks! I appreciate it.

Issue #1873, opened 2017-07-18
Create a Hint about drag/drop of tags


BjarniRunar:
It is not immediately obvious that dragging and dropping tags (or messages) is possible, nor is it obvious how the dragging behaves.

We should add a hint that tells the user about this feature.

To clarify, the rules are:

   * Dragging a message FROM a search result view ONTO a tag is "moving the message", which means any tags you are currently searching for get removed and the tag it's dropped on gets added.
   * Dragging a tag FROM the sidebar ONTO a message is "tagging", it just adds the tag.

Tagging is straightforward enough, but the moving behaviour is particularly tricky and needs explaining, because our current UI doesn't really make it super visually obvious what the current search context is.

For example, if you're looking at the search "in:inbox in:potato", (the intersection of messages in both Inbox and Potato), dragging a message onto the Salad tag will remove both Inbox and Potato, and add Salad. However, dragging onto Inbox will only remove Potato.

It would be nice if someone volunteered to do this before 1.0, but it's not a blocker. Hint: The hints code lives in `shared-data/contrib/hints/`

(thanks to @karlp for the feedback & discussion)

PR #2055, opened 2018-04-22
Dragging and dropping tags hint

kolibrik7:
I added a hint about dragging and dropping tags and messages (issue #1873).

### DIFF
--- a/shared-data/contrib/hints/hints/dragging-tags.html
+++ b/shared-data/contrib/hints/hints/dragging-tags.html
@@ -1,0 +1,24 @@
+{% extends "layouts/" + render_mode + "-tall.html" %}
+{% block title %}{{_("Dragging and dropping tags")}}{% endblock %}
+{% block content %}
+<div class="content-normal" style="max-width: 46em;">
+
+  <h1><span class="icon-tag"></span>
+    {{_("Dragging and dropping tags")}}
+  </h1>
+  <p>
+  	{{_("You can drag and drop tags or messages to add or change tags.")}}<br>
+  	{{_("The rules are following:")}}
+  </p>
+  <ul class="list" style="margin-left: 2em; list-style: disc;">
+  	<li>{{_("If you drag a message from a search result view onto a tag, all tags from the message are removed and the tag you drop the message on is added.")}}</li>
+  	<li>{{_("if you drag a tag from the sidebar onto a message, the tag is added to the message and other existing tags are not removed.")}}</li>
+  </ul>
+  <p>{{_("For example:")}}<br></p>
+  <ul class="list" style="margin-left: 2em; list-style: disc;">
+  	<li>{{_("If you're looking at the search 'in:inbox in:job', which is the intersection of messages in both Inbox and Job, dragging a message onto the Trash tag will remove both Inbox and Job tags and add the Trash tag.")}}</li>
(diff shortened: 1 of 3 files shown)


BjarniRunar:
Thanks, this is awesome!

There are some mistakes in your explanations though; would you like me to give you feedback and update the PR, or would you prefer I merge and just edit things to my liking myself? I'm fine with either way.

kolibrik7:
Hi,
I'd like to know what were the mistakes, but feel free to merge and edit what you want.

shared-data/contrib/hints/hints/dragging-tags.html
```diff
@@ -0,0 +1,24 @@
...
+
+  <h1><span class="icon-tag"></span>
+    {{_("Dragging and dropping tags")}}
+  </h1>
+  <p>
+  	{{_("You can drag and drop tags or messages to add or change tags.")}}<br>
+  	{{_("The rules are following:")}}
+  </p>
+  <ul class="list" style="margin-left: 2em; list-style: disc;">
+  	<li>{{_("If you drag a message from a search result view onto a tag, all tags from the message are removed and the tag you drop the message on is added.")}}</li>
```
BjarniRunar: This is not quite correct - not all tags are removed.

Only the tags you are currently looking at are removed. Dragging a message is a "move from this to that" action, where this might be "in:inbox in:new" and that might be "trash". If the messages were also tagged as "exciting", but "exciting" was not part of the search criteria, the "exciting" tag will persist after the move.

shared-data/contrib/hints/hints/dragging-tags.html
```diff
@@ -0,0 +1,24 @@
+{% extends "layouts/" + render_mode + "-tall.html" %}
+{% block title %}{{_("Dragging and dropping tags")}}{% endblock %}
+{% block content %}
+<div class="content-normal" style="max-width: 46em;">
+
+  <h1><span class="icon-tag"></span>
+    {{_("Dragging and dropping tags")}}
+  </h1>
+  <p>
```
BjarniRunar: Is it worth adding a few words to explain the tag concept itself here? Not everyone is familiar with the concept; using synonyms ("labels", "categories") and explaining how tags differ from mailboxes?

shared-data/contrib/hints/hints/dragging-tags.html
```diff
@@ -0,0 +1,24 @@
...
+  	{{_("You can drag and drop tags or messages to add or change tags.")}}<br>
+  	{{_("The rules are following:")}}
+  </p>
+  <ul class="list" style="margin-left: 2em; list-style: disc;">
+  	<li>{{_("If you drag a message from a search result view onto a tag, all tags from the message are removed and the tag you drop the message on is added.")}}</li>
+  	<li>{{_("if you drag a tag from the sidebar onto a message, the tag is added to the message and other existing tags are not removed.")}}</li>
```
BjarniRunar: I think it might be better to omit the word "existing" here - it's not the right word to describe the relationship between a tag and the messages it has been applied to. In general, the vocabulary here is tricky. Are tags added and removed? Are they applied? Are messages tagged and untagged? Sorry if this is nit-picky, but since tags are confusing to many people I think it's important that our descriptions are as clear as possible.

kolibrik7:
Thanks for feedback!
I don't have much experience with Mailpile, so I think it'd be better if you merge and edit things.

BjarniRunar:
This is merged in my local tree. Will push up soon. Thanks again for doing this!

Issue #1352, opened 2015-07-28
Allow to: from: search queries to match domain names


bnvk:
Currently, using the `to:name@email.com` and `from:name@email.com` search queries does not accept wildcard queries like `from:*@email.com` however, this would be really useful to bulk sort "all contact from a given company / org / domain into one search"

BjarniRunar:
Wildcards are never going to happen, because of the way the search index is built. It just can't do that.

However, making it possible to search from/to/... by domain is easy and accomplishes what you want. I'll rename the issue; I agree this would be very useful.

BjarniRunar:
I think this was fixed by #2057. Thanks @takano-akio !

PR #2057, opened 2018-04-24
Allow the domain name in e-mail addresses to be searched for (#1352)

takano-akio:
If an email contains 'From: foo@example.com', now it can be found by the search query 'from:example.com'.

I tried this change locally and it seems to work. However I don't really know the proper way to test it. I wasn't able to figure out how to run the test suite.


BjarniRunar:
Thanks for doing this! I agree this is something people want to be able to do explicitly.

takano-akio:
Thanks! Can #1352 be closed now?

BjarniRunar:
Looks like it! Thanks!

Issue #2044, opened 2018-04-12
Index of new mail source vulnerable to unclean shutdown


JackDca:
When a newly created Mailpile instance is set up, including a profile with an IMAP account, all the messages in the IMAP acount are indexed. This takes on the order of 1-2 s per message (e.g. a few minutes for 250 messages), and Mailpile does the indexing in batches (e.g. 10-100 messages). If there is no other activity (e.g. quit command, new mail on the IMAP account, user actions on the UI), the metadata index is not written to the `mailpile.idx` file after completion of the indexing, even after a wait of many hours. An unclean shutdown while in this state will result in a pile (workdir) with no `mailpile.idx` file. Nevertheless the keyword index file `kw-journal.dat` is written out.

When Mailpile is run a second time, the indexing operation is eventually repeated, apparently as a result of rescans of the mailbox, but it may be hours before all the messages are indexed so that they appear in the UI. Since `kw-journal.dat` has been written, ghost entries are created in the internal metadata index for all the `msg_mid`s that were created in the first run and new `msg_mid`s are created in the second indexing operation.

(This information verified using a shell script that records each change to the size or modification time of `mailpile.idx`.)

It is possible that similar behaviour occurs when a new profile with an IMAP account or other mail source is added to an existing Mailpile instance - this has not been tested.

It would be better if `mailpile.idx` could be updated as the messages from a new IMAP account (or any mail source) are indexed. However, if an append to `mailpile.idx` were done after each message, the result would be wasteful of space in `mailpile.idx` and also likely of CPU time. It appears that each append to mailpile.idx has an overhead of 300 - 400 bytes in the encrypted file and the CPU cost of handling multiple encrypted segments may be significant. In addition, after `MAX_INCREMENTAL_SAVES = 25` appends, Mailpile forces a complete rewrite of `mailpile.idx`. If a new IMAP account with hundreds of messages is added to an existing Mailpile with a large `mailpile.idx` then multiple complete rewrites of `mailpile.idx` will take place which is  expensive. A more efficient solution would be to append to `mailpile.idx` after completing each batch of messages.

BjarniRunar:
I think we need to expect tens of thousands of messages, not hundreds.

Hmm, I know you've done a bunch of testing. I'm surprised we don't write out for hours, I agree that is excessive. 

[...]

PR #2063, opened 2018-05-02
Save the metadata index more often

JackDca:
To reduce vulnerability to unclean shutdowns (#2044):
- Saves to `mailpile.idx` are done every 15 minutes by a cron task if there is any change to save. This is in addition to the saves previously implemented in the code, which are triggered by message tagging and similar events.
- Most saves are appends, not full rewrites.
- A full rewrite is performed if the total number of appends to the file exceeds 1/200 of the number of lines in the file (minimum 5000/200 = 25).
- A full rewrite of the file is performed if the number of duplicated lines in the file exceeds 1/10 of the number of lines in the file (minimum 5000/10 = 500). Duplicated lines are those which repeat a previous `MSG_MID` (for message entries) or Contact Index ID (for contact entries).
- Notwithstanding the above, a full rewrite is **never** performed when Mailpile is shutting down.
- Each append is introduced by a blank line (`"\n"`) in the plaintext `mailpile.idx`.
- When loading the file at Mailpile startup, the number of appends already in the file is initialized by counting blank lines, then that number is incremented whenever a new append is made.
- When loading the file at Mailpile startup, the number of lines already in the file is counted, then that count is updated whenever a new append is done. The difference between the number of lines in the file, and the combined number of items in the internal `INDEX` and `EMAILS` lists, is the number of duplicated lines.

The code was tested on a sample IMAP account with about 250 messages and on a production Mailpile instance with 50,000 messages by changing the parameter values (200, 10, 5000) that control full rewrites to exercise all parts of the new code.


BjarniRunar:
Cool, thanks for doing this!

I mostly like this approach, except for the blank-lines-with-meaning hack. That feels brittle to me. How about, on load, comparing a) how many lines were actually read from the file with b) how large the index really is (`len(self.INDEX)`) - that gives you a ratio that tells you roughly the same thing? It doesn't matter if the number is 100% accurate or not, it just needs to be "good enough" to prevent runaway index bloat.

OTOH - I *really* like your change to prevent a full rewrite on shutdown, that's a major improvement!

JackDca:
This PR already calculates a (total lines) - (`len(self.INDEX)`) difference ("duplicate" line count) and puts a limit on it. Problem is, if one email arrives in one 15 minute interval, it generates one append,  one new line and one INDEX entry. This leaves the "duplicate" count unchanged. Yet the execution time overhead to write/read one append is on the order of 20x the time required to write one line (see results and rationale in latest comments in #2044).

I could get rid of  the blank lines and the explicit append count, while lowering the permitted "duplicate" lines to 1% of total lines from the present 10%. In _normal_ use, each incoming email will eventually be read, trashed, and/or tagged, which _will_ create a "duplicate" line (and also another append!). So in _normal_ use the limit on the "duplicate" count will also roughly limit the number  of appends in the file.

The fragility you reference is as you explained in #1566, that is a future contributor might not be aware of that detail of the file format. One way to address that is by documentation, both internal and external. See [Data formats - Metadata index](https://github.com/mailpile/Mailpile/wiki/Data-formats:-Metadata-index) . My concern is code that is fragile in another way, i.e. it works well under _normal_  use but not so well in the edge cases. I like rigor.

In any case, I'll modify the PR as per your request. I should get to it on Monday.

JackDca:
@BjarniRunar - after letting this simmer for a while, I am not satisfied with the heuristic approach of limiting the number of duplicate lines in the `mailpile.idx` file without explicitly limiting the number of appends. Reducing the limit on duplicate lines to 1% will increase the number of complete rewrites unnecessarily. Agreed, the use of a blank line to mark the start of an append was a bit too clever and not obvious to the reader. So I'd like to mark the start of an append with `#-----APPENDED SECTION-----\n`.

The code has always recognized lines starting with # as comments so this change is backward compatible.  It affects the way the program handles the file without affecting the information content of the file; this is consistent with other use of comments to encode metadata e.g. hash-bang #! at the start of scripts which is not part of the script itself but helps the OS handle it; similar stuff is done in HTML.

If you can accept this approach I'll update the PR.

BjarniRunar:
Go for it. :+1:

JackDca:
Done.

This was tested thoroughly when the append marker was a blank line (see above). Just to be sure, accelerated testing was done on this commit by modifying the mailpile.idx write interval to 300 s from 900 s and ITEM_COUNT_OFFSET to 250 from 5000 for the duration of the test.

BjarniRunar:
Thanks! Merging this in my local tree so I can kick the tires a bit before pushing.

PR #2065, opened 2018-05-07
Remove BrowseOrLaunch

noahhaasis:
I want to resolve#1993 which states, that the browse_or_launch function is obsolute and should be removed.


BjarniRunar:
Cool! However, I think this is incomplete.

Isn't there a function at the bottom of the file that registers this class as an API method?

noahhaasis:
Thank you for your hint.

Issue #2061, opened 2018-05-01
Mailpile custom GnuPG keychain and gpg.conf


BjarniRunar:
One of the decisions from the little Mailpile dev summit last week, was that on Desktops, Mailpile should probably (by default) use a private keychain, rather than the shared OS keychain.

There is currently an option for this, but it is incomplete in that we will still be using the shared GnuPG configuration. Mailpile should use its own GnuPG configuration file.

Another issue we found, was that at build time, GnuPG 2.2.x on Mac seems to compile in the paths to dirmngr and gnupg-agent and other helper utilities. When the MacOS .app bundle gets moved around, these paths break and GnuPG stops working. I am pretty sure this could also be resolved by Mailpile shipping its own gpg.conf file.

Fractal complexitiy - this balloons to four possibilities:

   1. Shared GnuPG keychain, shared GnuPG config
   2. Shared GnuPG keychain, private GnuPG config
   3. Private GnuPG keychain, shared GnuPG config
   4. Private GnuPG keychain, private GnuPG config

Of these options, I see strong use-cases for 1. and 4. - case 2. may be required for the Mac bundle to work. I see no use for case 3.

If we can eliminate case 2, this becomes much simpler. That would imply requiring that IF the user wants a shared keychain, they must also provide a shared GnuPG installation that Mailpile can use instead of the one bundled in the .app. That makes shared keychains an expert-level setting. This needs a little more thought.

JackDca:
I agree with these comments. "Casual users" should use "case 4" by default so they don't have to understand the interaction between Mailpile and an external GnuPG. "Expert users" should be able to enable "case 1" if they wish.

The Windows installer build script in the current master installs the executables of GnuPG in a subdirectory of the Mailpile program directory: `C:\Program Files (x86)\Mailpile ehf\Mailpile|\GnuPG` on WIn7. The thinking was that (since standard Windows does not have a package manager) casual users should not be asked to manually install additional packages; the Mailpile install should be just one operation; and putting Mailpile's GnuPG executables in that location would prevent conflict with another GnuPG if it were installed. However, the installer by default still leaves the GnuPG data at a public location, `C:\Users\[username]\AppData\Roaming\gnupg` on Win7, which creates the possibility for conflict. The private location should be the default.

JackDca:
Re the comment: "Mailpile should use its own GnuPG configuration file."

I think this is already covered. According to the gpg manpage, if the command is
[...]

PR #2066, opened 2018-05-08
Windows installer updates

JackDca:
The existing Windows installer may become obsolete thanks to @AlexanderHaase 's work. To bring the existing installer to a tidy conclusion, this update fixes a couple of rough spots in it and brings the third party package references up to date. It may be useful to someone in the future (or even to Alex).
- In the installer, do not pop up command windows, but direct Python installer and pip output to the installer window.
- When building the Windows installer make sure it gets the git branch name for the version identifier from the Mailpile commit being packaged.
- Pass python-[version].msi file name to mailpile.nsi in an argument so all third party version updates can be made in repackage-linux.sh.
- Completely delete the previous Win installer WORKDIR when starting new build.
- Update third party package references to the latest versions.

Originally this update included a mod to mailpile/config/manager.py to set the default GnuPG data directory to the Mailpile data directory (on Windows installs only), consistent with the fact that this installer a sets up a private copy of the GnuPG executable in the Mailpile program directory. Since this default location may be adopted for all platforms per #2061, the mod was deleted from this PR. It can be reenabled if wanted by
`git revert 9d42716`


BjarniRunar:
Cool!  I feel bad that this is likely to get clobbered when we merge Alex's work, but merging all the same.

Issue #2045, opened 2018-04-16
Top right UI Buttons are rendered out of screen when the browser's size is short.


ghost:
UI Buttons disappear (are rendered out of screen) when narrowing a short browser window.
In order to reproduce this bug, it is imperative that the browser's window is short (low vertical height).
This bug was observed on macOS 10.13.4 using Safari 11.1.
I was able to reproduce it on the, now, latest versions of Firefox and Chrome.

I demonstrate the bug with a sequence of screenshots.

The following is a *normal* screenshot of Mailpile in a browser.
![screen shot 2018-04-16 at 09 31 41](https://user-images.githubusercontent.com/1230353/38795711-65d65abc-4159-11e8-9abf-0fba25c939b8.png)
The following three screenshots are *abnormal*, they show the result of gradually narrowing the browsers size. Note how the icons in the top right disappear one after another.
![screen shot 2018-04-16 at 09 31 45](https://user-images.githubusercontent.com/1230353/38795747-8258a03c-4159-11e8-8b04-48b57f50c39b.png)
![screen shot 2018-04-16 at 09 31 47](https://user-images.githubusercontent.com/1230353/38795755-8552cf38-4159-11e8-8040-dd95f699788c.png)
![screen shot 2018-04-16 at 09 31 53](https://user-images.githubusercontent.com/1230353/38795761-88e62bfe-4159-11e8-9bca-5066de2c1214.png)
The following screenshot is *normal*, it shows that when the window gets too narrow, the a different style is applied.
![screen shot 2018-04-16 at 09 31 56](https://user-images.githubusercontent.com/1230353/38795791-9e05567c-4159-11e8-8a22-3b6d88c7c91f.png)
I must add that altough the final screenshot is *normal*, I expected to see five icons -- not just three.

BjarniRunar:
The responsive UI needs more breakpoints, it seems. Or just be more aggressive about making the Report Bugs link disappear, I think it's what pushes everything else out of place.

The three instead of five thing is "working as intended", as the screen gets smaller we are forced to start dropping features. Someday someone will create or plug in an overflow menu handler that makes the dropped ones still accessible.

ahadmushir:
Hey! 
Is this issue still there? I would like to work on it as a starter of my open source community contribution :)

BjarniRunar:
Hi @ahadmushir , sorry about the late reply!

[...]

PR #2094, opened 2018-06-06
Fix rendering of top right UI buttons

kolibrik7:
This resolves issue #2045. I changed some things in topbar and added a new class, tablet-hide.


BjarniRunar:
This looks sane! Thanks.

If you do more of these, it would be really helpful if you could include before/after screenshots to show what you have changed. It's a little hard for me to review things if I have to render the CSS in my head.

PR #2101, opened 2018-06-16
Fix broken « report a bug » display

JocelynDelalande:
That is not more or less elegant CSS than before, but it fixes display.

before:

image

after:

image


BjarniRunar:
This is going to break the layout on smaller screens again. I don't know if we care, since the Report Bug link is only supposed to be present in developer builds anyway. So... merging.

Thank you!

Issue #2031, opened 2018-04-05
Implement "Factory Reset" for user tags


BjarniRunar:
Sometimes users move tags around and they get lost. This is disconcerting and can be hard to recover from for people not deeply familiar with the innards of the app.

Rather than make it harder to customize the UI, I think it might make sense to add a "Factory Reset" feature to the sidebar, which reverts all of the tags that are initially configured on setup to their original icons, colors, positions, etc. etc.

How I would do this:

   1) Add a `reset/tags` API method to `mailpile.plugins.setup_magic` - some refactoring will be required
   2) Expose this method to the user as a subtle option when they click `Organize` in the sidebar.

Regarding 2, we don't want people to accidentally revert their tag settings, so there should definitely be a warning popup that asks the user to confirm. A javascript `confirm()` would probably suffice.

(A more fancy implementation would allow the user to select exactly which of the standard tags get reset, but that's a lot of extra work which may not be justified at this point in time.)

SushantGahlot:
Hi, I am new to contributing to OSS and I'd like to work on this if nobody has started working on it?

BjarniRunar:
Hi @SushantGahlot and welcome! However, @kolibrik7 had already start working on this issue. You can see the pull request linked above your comment; unless they have given up on completing that it would probably be best to find something else to work on.

PR #2103, opened 2018-06-18
"Factory Reset" for user tags

kolibrik7:
This PR implements "Factory Reset" for user tags, it works basically as described in #2031 .

I added a new "default" attribute to tags that are initially configured on setup, so that we know what was the original tag.

### DIFF
--- a/mailpile/plugins/setup_magic.py
+++ b/mailpile/plugins/setup_magic.py
@@ -22,7 +22,7 @@ from mailpile.plugins.contacts import ListProfiles
 from mailpile.plugins.migrate import Migrate
 from mailpile.plugins.motd import MOTD_URL_TOR_ONLY_NO_MARS
 from mailpile.plugins.setup_magic_ispdb import STATIC_ISPDB
-from mailpile.plugins.tags import AddTag
+from mailpile.plugins.tags import AddTag, GetTags
 from mailpile.commands import Command
 from mailpile.crypto.gpgi import SignatureInfo, EncryptionInfo
 from mailpile.eventlog import Event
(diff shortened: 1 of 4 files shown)


BjarniRunar:
Hello! Thank you so much for working on this.

Sorry it's taken me a few days to respond, but I have a couple of concerns about the implementation and I wanted to take the time to give good feedback. My concerns:

Most pressingly, your `ResetTags.setup_command` is duplicating all sorts of settings from the dict() in `SetupMagic`. I feel quite strongly that the same data should never exist in two places, because that becomes a trap for future developers - if things get changed in one place but not the other later on, then we end up with nasty, sneaky bugs. Both commands should use the same data source for their work; since `SetupMagic` is an ancestor of your `ResetTags`, you could just access it directly as `self.TAGS`, right?

Secondly: are you sure it is necessary to add the `default` attribute to the tags? This adds a field to every single tag in the user's configuration; if this can be done in a less bloated way that would be preferable. It seems to me like a simple list in ResetTags declaring which tags are affected by the factory reset would suffice. Do you agree?

Thirdly: Please don't add UI elements using Javascript. For consistency with the rest of the app they should be declared in the HTML template along with the other buttons. The template that needs changing is `..../shared-data/default-theme/html/partials/sidebar.html`, grep for `Done`.

...

Sorry I had so many complaints, but keeping the code maintainable is quite important.

It does look like what you've written will work (given the current set of tags); I think the only feature request I would have would be to present the user with a modal dialog instead of `comfirm()` that lets them choose which tags to reset. But that's a stretch goal (maybe a 2nd iteration), I would definitely not reject the PR for lack of that.

BjarniRunar:
Note to readers: This isn't getting merged as is, and the conversation has stalled. If anyone else wants to pick this up and implement my requests that would be super awesome. Otherwise I might do it myself someday, but I have bigger fish to fry so it might take a while.

PR #2104, opened 2018-06-20
Updated build system references to be up-to-date with available third-party sources.

AlexanderHaase:
Tor project changed the URL for tor binaries(checksum intact)
Resource hacker released a new major version. Sadly they don't have historic release archives. We could probably script resource modification via pywin32 if resource hacker proves too unstable.

Also updated platform.py to expect binaries to exist on PATH on all platforms.


BjarniRunar:
Thank you!

PR #2105, opened 2018-06-20
Fix TLS Certificate tool dialog with non-root location

foursixnine:
Hi, while trying to setup Mailpile, found this small problem :)

Let me know if I need to provide something else aside from the commit


BjarniRunar:
Hello!

Prefixing the path is exactly what the U() function does! Your change would prefix the http_path twice! :wink:  So I am pretty sure this isn't the fix to whatever your problem is...

If you could reproduce and document the problem you are having in an issue, that would be good?

foursixnine:
@BjarniRunar Can do :)

PR #2109, opened 2018-06-22
Fix spurious decryption failures (#2093) caused by GnuPG 2.1 message changes

### LINKED ISSUE #2093: Decryption failure after startup
My production Mailpile, when just started up, initially fails to decrypt most emails that have been successfully decrypted before.

The code is master commit a12b493 dated 2018-05-10 with some development mods which should not affect this issue. The system is Debian stretch with GnuPG 2.1.18 installed.

The encryption key is my main email key, predating Mailpile. It is "unlocked, password remembered" according to the Encryption Key Management popup and confirmed as follows:   
`mailpile> print secrets/3db19e0dde466a2e3e0c976d0af4af7188260f5d`   
`Elapsed: 0.001s (print: Displayed settings)`
`{`
`    "secrets/3db19e0dde466a2e3e0c976d0af4af7188260f5d": {`
`        "password": "(SUPPRESSED)", `
`        "policy": "store" `
`    }`
`}`   

Digging into this, the problem seems to be in `mailpile.crypto.gpgi.GnuPG.decrypt()`. That method is called with no key or passphrase specified. The code is designed to make two decryption attempts. In the first attempt the passphrase most recently used by earlier calls to `mailpile.crypto.gpgi.py.GnuPG` methods is passed to gpg. If decryption is unsuccessful, the output of `gpg` is parsed to find a line containing the term `NEED_PASSPHRASE` which identifies the encryption key that is needed. Then the passphrase for that key is retrieved and is sent to gpg for a second decryption attempt.

Adding debug output shows that the most recently used passphrase is the Mailpile passphrase (the one used to authenticate on the web interface), probably because it has just been used to decrypt the Mailpile master key, so it does not work with the needed encryption key. Then, the search for `NEED_PASSPHRASE` fails because gpg does not output such a line, although the needed information is present in the lines that gpg does output on `syserr` and `status`.

Testing gpg  in standalone mode (from the bash command line), starting with the identical arguments that are passed to it by Mailpile in the decryption attempt, with the encrypted message input from a file, shows that gpg outputs the `NEED_PASSPHRASE` line only if both the `--batch` and `--passphrase-fd=0` options are deleted. (Possibly this is a change from GnuPG 1.4?)

Possible fixes:   
1. Delete the `--batch` and `--passphrase-fd=0` options from the arguments passed to `gpg`. This is fraught with peril because `gpg` will behave as if it is in interactive mode rather than batch mode, which may have unexpected effects on the many other calls to gpg from Mailpile.
2. Get the information about the required key from the existing outputs of gpg (to `stderr` and `status`). Doable but a bit tricky because no one output line clearly identifies the primary key id of the one passphrase that is needed.
3. Call gpg twice, once to get the set of "encrypted to" keys, a second time to decrypt. The set of "encrypted to" keys would be AND'd with the Mailpile's set of available private keys and the passphrase of the needed key would be sent to `gpg`.
4. Like 3, but enhance `mailpile.crypto.gpgi.py.GnuPG.sniff()`, which is used to identify data that has PGP content without calling gpg, to get the set of"encrypted to" keys, in place of the first call to `gpg`.

Also, the most recently used passphrase is sometimes not a private key passphrase but something that can't possibly be the needed passphrase (as in the test described). It might be a better guess, in the first decryption attempt, to send the passphrase of the private key associated with the receiving email account.

Here's a sample of the output of gpg, captured on the unsuccessful decryption attempt.
(The keys mentioned here are available on public keyservers but I have hidden the email addresses.)

`<<STATUS<< [GNUPG:] PROGRESS stdin ? 0 0 B`
`<<STDERR<< gpg: armor header: Version: Hush 3.0`
`<<STDERR<< gpg: public key is 5CB0ABB0E137C4FB`
`<<STATUS<< [GNUPG:] ENC_TO 5CB0ABB0E137C4FB 1 0`
`<<STDERR<< gpg: using subkey 5CB0ABB0E137C4FB instead of primary key 0AF4AF7188260F5D`
`<<STATUS<< [GNUPG:] KEY_CONSIDERED 3DB19E0DDE466A2E3E0C976D0AF4AF7188260F5D 0`
`<<STDERR<< gpg: public key is 1F393109F3CAB720`
`<<STATUS<< [GNUPG:] ENC_TO 1F393109F3CAB720 2 0`
`<<STDERR<< gpg: using subkey 1F393109F3CAB720 instead of primary key CDF01E045BC3E0CC`
`<<STATUS<< [GNUPG:] KEY_CONSIDERED FB1CDEA8F2416FB2C7F940BDCDF01E045BC3E0CC 0`
`<<STDERR<< gpg: encrypted with 2048-bit RSA key, ID 1F393109F3CAB720, created 2008-07-11`
`<<STDERR<<       ""sender@****" <sender@****>"`
`<<STATUS<< [GNUPG:] NO_SECKEY 1F393109F3CAB720`
`<<STDERR<< gpg: using subkey 5CB0ABB0E137C4FB instead of primary key 0AF4AF7188260F5D`
`<<STATUS<< [GNUPG:] KEY_CONSIDERED 3DB19E0DDE466A2E3E0C976D0AF4AF7188260F5D 0`
`<<STDERR<< gpg: encrypted with 2048-bit RSA key, ID 5CB0ABB0E137C4FB, created 2007-01-21`
`<<STDERR<<       ""me-recipient@****" <me-recipient@****>"`
`<<STDERR<< gpg: public key decryption failed: Bad passphrase`
`<<STATUS<< [GNUPG:] ERROR pkdecrypt_failed 67108875`
`<<STATUS<< [GNUPG:] BEGIN_DECRYPTION`
`<<STATUS<< [GNUPG:] DECRYPTION_FAILED`
`<<STDERR<< gpg: decryption failed: No secret key`
`<<STATUS<< [GNUPG:] PROGRESS stdin ? 3616 0 B`
`<<STATUS<< [GNUPG:] END_DECRYPTION`

JackDca:
This code modifes `mailpile.crypto.gpgi.GnuPG.decrypt()` so that it works with `gpg` 2.1.

In `gpg` 1.4, if a secret key passphrase is needed, a `NEED_PASSPHRASE` message appears in the `gpg` output. In `gpg` 2.1 this message does not appear in the `gpg` output when the `--batch` or `--passphrase-fd=0` options are used. This code adds a check for `KEY_CONSIDERED` messages (which are output by 2.1, but not 1.4) to identify the needed key and passphrase.

The original code also checked the `gpg` return code (`retvals[0] == 0`) to determine if the decryption was successful. This caused duplicated calls to `gpg` to decrypt (even with gpg 1.4) in cases where the decryption succeeds but the return code is non-zero. Example: when decryption is successful but the message is signed by an untrusted key. This code instead checks for a `DECRYPTION_OKAY` message.

This fixes #2093. It has been tested with both `gpg` 1.4.21 and 2.1.18 under Debian Stretch.


BjarniRunar:
That looks great, thank you!

PR #2113, opened 2018-07-04
Treat [new] encrypted messages without MDC as broken

stoksc:


BjarniRunar:
Oh! I just realized, there is one bug in this: you aren't checking the `decrypt_requires_MDC` parameter. 

So after this change it becomes impossible to decrypt these messages, even if the user has configured Mailpile to allow it. Check the logic around the `DECRYPTION_OKAY` keyword handling for an example of what I mean.

BjarniRunar:
Closing this; it's been superceded by another PR which got merged. Thanks for working on this though, I think this PR inspired the other one.

stoksc:
Thanks for the thanks! I started a new job and got terribly busy. Glad it got handled.

PR #2115, opened 2018-07-07
Add pysocks dependency to debian pkg

JocelynDelalande:
Seems to have been forgotten in 03ebb435b5e97ae15a8f8f79ba015beac1a465b5

I also updated the wiki https://github.com/mailpile/Mailpile/wiki/Getting-started-on-Linux

(maybe Debian is not enough and there are other pkgs to be updated)

### DIFF
--- a/packages/debian/control
+++ b/packages/debian/control
@@ -61,6 +61,7 @@ Depends: ${misc:Depends},
          python-imgsize,
          python-socksipychain,
          python-dns,
+         python-socks,
          python-pgpdump,
          python-pil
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Thanks! However, this is not needed - pysocksipychain is there and does the job. The reason I'm not using it in the Python requirements.txt file, is I haven't pushed it to PyPI yet.

BjarniRunar:
As long as you don't mind me closing the PR, this is an excellent way to have a conversation! ;-)

PR #2120, opened 2018-07-20
Recognize vfs aliases /Mailpile$, /Share$, /Home$ in cd command.

JackDca:
The CLI ls (browse) command and other commands recognize the aliases created by Maipile's VFS. For some reason cd does not. This two line patch implements the aliases in cd.


BjarniRunar:
Thanks!

PR #2125, opened 2018-07-26
Check wether url prefix is fully present

### LINKED ISSUE #2106: U function with non-root location does not add http_path to /crypto/tls/getcert/ route
So while trying to use latest release and master from mailpile, I get into trouble when trying to use the    TLS Certificate Tool. Applying:  changes in  #2105 solved my problem.

When clicking the button, if aforementioned changes are not applied, `AM: Error: error` is shown in the JS console, `AM: Success!` is shown when changes are applied, and modal window shows up as expected. I used [this guide](https://github.com/mailpile/Mailpile/wiki/Accessing-The-GUI-Over-Internet#using-nginx-as-proxy) to set it up. Everything else seems to work without problems

Mailpile version: v1.0.0rc3 (Current master)

Nginx configuration for that host:
```
  location /c {
    access_log /var/log/nginx/mailpile_access.log;
    error_log /var/log/nginx/mailpile_error.log info;

    proxy_pass http://127.0.0.1:33411;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-Server $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  }
```

Mailpile sys.http_path

```
mailpile> P sys.http_path
Elapsed: 0.001s (print: Displayed settings)

{
    "sys.http_path": "/c"
}

```
Screenshot without the changes:

2018-06-20-230015_1065x745_scrot

With the changes everything works.

foursixnine:
When using 'sys.http_path' as '/c' whenever the certificate tool dialog
needs to be brought up, the U function gets confused and can lead to
certain urls not working, therefore: check the full prefix instead, to
avoid other combinations to cause the same problem

Fixes #2106

### DIFF
--- a/mailpile/www/jinjaextensions.py
+++ b/mailpile/www/jinjaextensions.py
@@ -741,7 +741,7 @@ class MailpileCommand(Extension):
         url = ''.join([unicode(p) for p in urlparts])
         if url[:1] in ('/', ):
             http_path = self.env.session.config.sys.http_path or ''
-            if not url.startswith(http_path):
+            if not url.startswith(http_path+'/'):
                 url = http_path + url
         return self._safe(url)
 
--- a/shared-data/default-theme/html/jsapi/index.js
+++ b/shared-data/default-theme/html/jsapi/index.js
@@ -353,7 +353,7 @@ Mailpile.API = {
 
   U: function(original_url) {
     var prefix = "{{ config.sys.http_path }}";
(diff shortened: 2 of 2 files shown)


BjarniRunar:
Thanks for doing this!

foursixnine:
You're welcome! :) Hope to keep giving a hand. Have fun!

PR #2126, opened 2018-07-27
Improved Tor/stem UX and pythonw compatibility

AlexanderHaase:


BjarniRunar:
Thanks for this! I've merged in my local tree, will push back up with my next batch of changes.

PR #2132, opened 2018-08-10
Add support for retrieving keys using Web Key Directory

### LINKED ISSUE #2117: Add support for GPG key lookup using Web Key Directory protocol
Hello,

Recent versions of GnuPG 2 ship with a new key discovery protocol - [WKD](https://wiki.gnupg.org/WKD), that is enabled by default. This scheme takes an e-mail address, converts it to HTTPS URL and downloads the binary key from that URL.

I think it would be a good fit for Mailpile as it already has an ability to discover keys (using local keyring or keyservers).

WKD has a benefit of always using secure connection (HTTPS) and contacting user's domain directly (on keyservers anyone can upload a key claiming to be anyone else).

 For example given this key:

```
$ gpg --with-wkd-hash -k team@mailpile.is
pub   rsa4096/0xAAD965A4707775F9 2013-11-29 [SC]
      Key fingerprint = 4161 CD18 CCA1 1A0B 1F4C  3EA0 AAD9 65A4 7077 75F9
uid                   [ unknown] Mailpile Team <team@mailpile.is>
                      4jeaxzyuxh4htn6eb5rc8977sf7i7qdu@mailpile.is
sub   rsa4096/0x2A9C9184BE71497E 2013-11-29 [E]
```

Then `gpg --locate-key team@mailpile.is` would try to download the binary key (*not* armored) from this URL: `https://mailpile.is/.well-known/openpgpkey/hu/4jeaxzyuxh4htn6eb5rc8977sf7i7qdu`. This of course wouldn't work now but for example [kernel.org supports WKD](https://www.kernel.org/category/signatures.html#using-the-web-key-directory) so `gpg --locate-key torvalds@kernel.org` would work.

One potential way to do this would be a new crypto key lookup handler (registered in `plugins/keylookup/__init__.py`).

I could work on a PR if this sounds like something in scope of Mailpile.

wiktor-k:
This change allows looking up keys using Web Key Directory scheme.
Functions for parsing key data have been extracted from email_keylookup.py into crypto/keydata.py as WKD also uses them for parsing keys.

Fixes #2117.

This is how it looks like when searching for `torvalds@kernel.org`:

screenshot from 2018-08-10 14-21-10

(I also maintain an address `test-wkd@metacode.biz` that can be found only using WKD).

I would very much appreciate comments about style (as I don't write Python) and suggestions how to mock HTTP calls for unit tests (all other lookup mechanisms have tests). Various constants (score, timeout, priority) are of course also to be discussed.

### DIFF
--- a/mailpile/crypto/keydata.py
+++ b/mailpile/crypto/keydata.py
@@ -1,0 +1,54 @@
+from pgpdump.utils import PgpdumpException
+import time
+import pgpdump
+
+def _get_creation_time(m):
+    """Compatibility shim, for differing versions of pgpdump"""
+    try:
+        return m.creation_time
+    except AttributeError:
+        try:
+            return m.datetime
+        except AttributeError:
+            return datetime.datetime(1970, 1, 1, 00, 00, 00)
+
+
+def get_keydata(data):
+    results = []
+    try:
+        if "-----BEGIN" in data:
+            ak = pgpdump.AsciiData(data)
+        else:
+            ak = pgpdump.BinaryData(data)
+        packets = list(ak.packets())
+    except (TypeError, IndexError, PgpdumpException):
+        return []
+
+    now = time.time()
+    for m in packets:
+        try:
+            if isinstance(m, pgpdump.packet.PublicKeyPacket):
+                size = str(int(1.024 *
+                               round(len('%x' % (m.modulus or 0)) / 0.256)))
+                validity = ('e'
+                            if (0 < (int(m.expiration_time or 0)) < now)
+                            else '')
+                results.append({
+                    "fingerprint": m.fingerprint,
+                    "created": _get_creation_time(m),
+                    "validity": validity,
+                    "keytype_name": (m.pub_algorithm or '').split()[0],
+                    "keysize": size,
+                    "uids": [],
+                })
+            if isinstance(m, pgpdump.packet.UserIDPacket) and results:
+                # FIXME: This used to happen with results=[], does that imply
+                #        UIDs sometimes come before the PublicKeyPacket?
+                results[-1]["uids"].append({"name": m.user_name,
+                                            "email": m.user_email})
+        except (TypeError, AttributeError, KeyError, IndexError, NameError):
+            import traceback
+            traceback.print_exc()
+
+    # This will only return keys that have UIDs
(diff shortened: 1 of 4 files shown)


BjarniRunar:
Thank you for doing this! I'll take a look as soon as I have some free time. :+1:

wiktor-k:
Hi @BjarniRunar, I'm glad I could contribute to Mailpile!

By the way I like the multi-layered approach to security in Mailpile (Tor, automatic generation of gpg keys, multiple key lookup schemes with scores), really interesting stuff! I wonder why others don't do that :smile:

wiktor-k:
Hi @BjarniRunar, sorry to bug you but is there any progress on the review? I'm afraid of bitrot for this PR code :)

wiktor-k:
Thanks @BjarniRunar!

BjarniRunar:
I've merged this in my local branch, I'm working on a bunch of other PGP related things right now so I'll just add this to the mix and test along with the rest of the things I'm doing. I'll post any comments back here, but if anything bugs me I'll just fix it at my end. Thanks again!

BjarniRunar:
Just FYI, since you requested feedback: I am very happy with this code! The only change I made was adding an exception handler so TLS certificate errors wouldn't cause noise.

When I push, this will become Mailpile's preferred remote key discovery handler, we won't hit keyservers at all if WKD gives useful results. Great stuff, thanks again!

wiktor-k:
Thanks a lot! I'm looking forward to your other PGP related things too, as I've said previously, Mailpile has a lot of good design decisions w.r.t. UX.

> When I push, this will become Mailpile's preferred remote key discovery handler, we won't hit keyservers at all if WKD gives useful results.

Glad to hear that! As far as I can see [GnuPG will also slowly move to use WKD exclusively for key lookup](https://lists.gnupg.org/pipermail/gnupg-users/2018-October/061034.html). (There are some ideas to [strip keyservers from UIDs](https://lists.gnupg.org/pipermail/gnupg-devel/2018-October/033969.html) - you know, GDPR - and have full keys in WKD only).

By the way should I close this PR manually?

BjarniRunar:
It should close automatically when the merge gets pushed to github.

wiktor-k:
Got it, thanks!

Issue #2127, opened 2018-07-29
Python Errors in Mailpile cli


TmCTux:
Hello,

here is the issue observed:
when entered "m", "mail" or as bellow "mail test", it's probably not a good command and use of the cli but for someone starting using mailpile it is confusing to have those error reported directly in the interface.
«
mailpile> mail test
What message is test?
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/mailpile/commands.py", line 584, in _run_sync
    return self._finishing(command(self, *args, **kwargs))
  File "/usr/lib/python2.7/dist-packages/mailpile/commands.py", line 582, in command
    return self.command(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/mailpile/plugins/compose.py", line 1035, in command
    return Update.command(self, create=create, outbox=outbox)
  File "/usr/lib/python2.7/dist-packages/mailpile/plugins/compose.py", line 999, in command
    email.update_from_string(session, update_string, final=outbox)
AttributeError: 'NoneType' object has no attribute 'update_from_string'

mail error: Failed: mail test
Elapsed: 0.029s (mail: Failed: mail test)

Aucun résultat
»

just interface improvement I guess.
thanks

nik1082:
Hello! Is this issue still up? I'll try and work on it!

This is my first time contributing to an open-source repo, so bear with me! I've already written a short solution to this problem, however I could not test it. Running on Windows 7 Ultimate, both Python 3.7 and 2.7 are installed. I already made sure that the virtualenv is running Python 2.7 and followed all the instructions on your wiki, however, running `start .\mp.cmd` still yields the following errors:

```
(MP-VIR~1) C:\...\Mailpile>Traceback (most recent call last):
  File "scripts\mailpile", line 11, in <module>
    from mailpile.app import Main
  File "C:\...\Mailpile\mailpile\__init__.py", line 1, in <module>
    from mailpile.i18n import gettext as _
  File "C:\...\Mailpile\mailpile\i18n.py", line 5, in <module>
    from jinja2 import Environment, BaseLoader, TemplateNotFound
ImportError: No module named jinja2
```

To verify that the dependencies are installed, here is my `pip list`:

```
Package                       Version
----------------------------- ---------
appdirs                       1.4.3
asn1crypto                    0.24.0
astroid                       1.6.5
backports.functools-lru-cache 1.5
certifi                       2018.8.13
cffi                          1.11.5
colorama                      0.3.9
[...]

PR #2133, opened 2018-08-14
Fix CLI traceback

nik1082:
Fix for issue #2127 

Upon entering the command 'mail test' in the CLI, this was the _old_ output:

```
mailpile> mail test
What message is test?
File "/usr/lib/python2.7/dist-packages/mailpile/commands.py", line 584, in _run_sync
return self._finishing(command(self, *args, **kwargs))
File "/usr/lib/python2.7/dist-packages/mailpile/commands.py", line 582, in command
return self.command(*args, **kwargs)
File "/usr/lib/python2.7/dist-packages/mailpile/plugins/compose.py", line 1035, in command
return Update.command(self, create=create, outbox=outbox)
File "/usr/lib/python2.7/dist-packages/mailpile/plugins/compose.py", line 999, in command
email.update_from_string(session, update_string, final=outbox)
AttributeError: 'NoneType' object has no attribute 'update_from_string'

mail error: Failed: mail test
Elapsed: 0.029s (mail: Failed: mail test)

No results
```

Fixed output:

```
mailpile> mail test
mail error: Cannot find message
Elapsed: 0.000s (mail: Cannot find message)

No results
```

### DIFF
--- a/mailpile/plugins/compose.py
+++ b/mailpile/plugins/compose.py
@@ -996,11 +996,11 @@ class Update(CompositionCommand):
                 if not Attach(session, data=self.data).command(emails=emails):
                     return self._error(_('Failed to attach files'))
 
-            try:
-                for email, update_string in email_updates:
-                    email.update_from_string(session, update_string, final=outbox)
-            except AttributeError:
-                return self._error(_('Cannot find message'))
+            for email, update_string in email_updates:
+                if not email:
+                    return self._error(_('Cannot find message'))
+                    break
+                email.update_from_string(session, update_string, final=outbox)
 
             emails = [e for e, u in email_updates]
             message = _('%d message(s) updated') % len(email_updates)


BjarniRunar:
Thanks for doing this! I have a couple of comments, are you open to making some changes? If not I'll just merge and tweak at my end, this is an improvement for sure.

Basically, I'd like this to be a bit more focused - currently, if for some reason there are multiple messages being updated (technically possible), we can end up with an inconsistent outcome where some message are updated and some are not, depending on where the loop triggers an exception.

Rather than wrap the loop in a try/except, it would be better to first check the list for any `email is None`, set the error message and skip the update loop entirely if we're in a bad state. Does that make sense?

nik1082:
Got the first part, but what do you mean by skip the update loop?

BjarniRunar:
I mean:

    if everything_is_ok:
        for things in things:
            ...

 :-)

BjarniRunar:
This is a definite improvement, thank you!

Perfect would have been one loop to check, followed by another loop to actually do things if the checks pass. As is, you still have the issue that some tasks can complete while others fail, depending on how far through the loop we get. But that should be a rare edge case and this is a clear improvement - so merging!

Issue #2137, opened 2018-09-06
Making an Autocrypt key from any OpenPGP key


JackDca:
There are good reasons for a user who already has an OpenPGP key to use that same key when they transition to Autocrypt. Most important, the user's contacts will already have some degree of trust in that key either through signed email history, TOFU, direct verification and/or web of trust. Also, in a group of users who have already exchanged keys, and are transitioning one by one to Autocrypt, the non-transitioned members will at all times retain the ability to verify the signatures of the transitioned members; this would not be the case if new keys were adopted.

There is an issue in using an existing OpenPGP key with Autocrypt. The Spec (Autocrypt Level 1 specification, Release 1.0.0) section 2.1.1 requires that the `keydata` in an Autocrypt header MUST have exactly five packets: primary key, user id, user id self-signature, encryption subkey, and subkey self-signature. The keys generated by Mailpile comply with this requirement. But, many existing OpenPGP keys have multiple user ids, encryption subkeys, signatures and other packets, making them non-compliant. This is especially true for users who have taken the web of trust seriously and obtained user id signatures from others.

This issue can be addressed by stripping an existing OpenPGP key of all but the required five packets to create a minimal, Autocrypt-compliant key. This appears to be feasable based on reading RFC4880. That key will always have the same primary key packet as the source OpenPGP key, but the other packets may depend on the selection criteria used. If there are multiple Autocrypt-enabled accounts with different email addresses (i.e. user ids), the same OpenPGP key could be the source for multiple Autocrypt-compliant keys.

GnuPG indexes keys based on the primary key packet. If GnuPG imports a key that has the same primary key packet as a key that is already in the keychain, then GnuPG merges the new key with the existing one. GnuPG will not import the proposed minimal key as a separate key. This limitation could be handled by having a separate GnuPG keychain for each Autocrypt-enabled account, but that would add complexity.

The alternative proposed here is to use one keychain, and to extract the minimal key from the source OpenPGP key each time it is needed to construct an Autocrypt header, using the email address associated with the account (the Autocrypt `from-addr`) as the selector. In effect, the Autocrypt `accounts[from-addr]` structure (Spec 1.3.2) would use the primary key fingerprint and `from-addr` to point to the `keydata`, which would actually be stored in the GnuPG keychain.

[...]

PR #2138, opened 2018-09-06
Making an Autocrypt key from any OpenPGP key - proof of concept

JackDca:
This is intended as the minimum code needed to try out the extraction of a minimal, Autocrypt-compliant key from any OpenPGP key as proposed in #2137. It is an Easter egg and not intended for "production" use. Hopefully by publishing it this way others will try the code.

The extraction procedure accepts an OpenPGP key (armored or binary) and returns a minimal PGP key     containing exactly five packets (base64 or binary) defining a primary key, a single user id with one self-signature, and a single encryption subkey with one self-signature. The unrevoked user id with newest unexpired self-signature and the unrevoked encryption-capable subkey with newest unexpired self-signature are selected from the input key. If `user_id` is specified by the calling code, a user id packet containing that string will be selected if there is one, otherwise any user id will be accepted. (The Autocrypt Spec states in sec. 2.1.1 "The content of the user id packet is only decorative.") If `subkey_id` is specified, only an encryption subkey with that id will be selected.

The procedure can be run stand-alone. The code prompts for the name of an OpenPGP public key file (binary or armored) and writes out the Autocrypt-compliant minimal key (binary) with .min.gpg appended to the input file name. Example:

`$ cd [Mailpile program root directory]`
`$ python mailpile/crypto/autocrypt_utils.py`

The `gpg --list-packets` command can be used to inspect and compare the input and output files.

The code can also be enabled in Mailpile to inject Autocrypt headers into outgoing emails by adding a line to the vCard for the email address associated with a Mailpile account. For example, if the account's email address is mptestdave@gmail.com, the vCard line can be added using the Mailpile CLI (after backing up the pile directory - _it's really easy to accidentally destroy the entire vCard!!_):

`mailpile> vcards/addlines mptestdave@gmail.com x-mailpile-autocrypt-accounts: enabled:true+prefer_encrypt:nopreference` (that's all one line)

Note that the modified vCard is saved to disk.

The modified vCard can be listed using the Mailpile CLI:

`mailpile> vcards --lines mptestdave@gmail.com`

That listing shows a line number for each vCard line. The line can be removed, thereby disabling the header injection code, using the Mailpile CLI (assuming the added line is line 8):

`mailpile> vcards/rmlines mptestdave@gmail.com 8`

### DIFF
--- a/mailpile/crypto/autocrypt_utils.py
+++ b/mailpile/crypto/autocrypt_utils.py
@@ -4,198 +4,279 @@
 
 import pgpdump
 import time, datetime
+import base64
+import textwrap
 
-class MinimalKey:
+def minimal_key(keydata, user_id = None, subkey_id = None,
+                    binary_out = False ):
 
-    def minimal_key(self, keydata, user_id = None, subkey_id = None ):
+    """
+    Accepts a PGP key (armored or binary) and returns a minimal PGP key
+    containing exactly five packets (base64 or binary) defining a
+    primary key, a single user id with one self-signature, and a 
+    single encryption subkey with one self-signature. Such a five packet
+    key MUST be used in Autocrypt headers (Level 1 Spec section 2.1.1). 
+    The unrevoked user id with newest unexpired self-signature and the
+    unrevoked encryption-capable subkey with newest unexpired
+    self-signature are selected from the input key.
+    If user_id is provided, a user id containing that string will be 
+    selected if there is one, otherwise any user id will be accepted.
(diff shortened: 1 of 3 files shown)


BjarniRunar:
Wow, cool. Sorry I've been AWOL and unresponsive.

I'm going to merge this, especially the key stripping code looks useful. Not sure about the VCard thing, there are other approaches to Autocrypt that are in flight (@smari did some work this summer that I need to merge) ... I need to put my coordinator hat on!

PR #2143, opened 2018-09-19
Change MSI GUIDs to use UUID v5 namespace scheme rather than UUID v4 random ids.

AlexanderHaase:
Eliminates the need to carry forward a UUID database across versions since they are now derived from the cryptographic hash of the destination path stub and underlying file. Changing file contents for the same destination path automatically causes the UUID to change whereas unchanged files retain the same UUID, exactly the desired update behavior.


BjarniRunar:
Thanks!

Issue #2142, opened 2018-09-14
Mailpile is trying to use a revoked key when setting up a new account


bjmgeek:
I am trying to add an account.  I've set the preference `Use shared GnuPG keychain for PGP encryption keys.` to "on".  When I get to the Encryption Key step, I see a pull down with my old (revoked) 1024 bit DSA key, but not my current one.   I've verified that the revocation certificate is in my gpg1 keychain, `~/.gnupg/pubring.gpg`. The current unrevoked key is also in my gpg1 keychain.  This is running on Debian unstable on amd64.

bjmgeek:
I noticed #1133, so I should mention that I first tested mailpile without gnupg1 installed, so `gpg` was version 2.2.10.  I then tried installing gnupg1, but got the same result.

JackDca:
PR #2142 should stop the use of disabled, revoked, or expired keys.

There is still the separate issue, that only one key is offered to the user, even though more than one suitable existing key is available. This is a separate issue.

bjmgeek - if it is not too difficult, it would be good if you could test this in the configuration that produced your original issue.

bjmgeek:
That pull request didn't work for me.  When I tried it, I cleared out my ~/.local/share/Mailpile directory, so I'd have a clean slate.  I started mailpile, and it asked me for a password, but then never progressed past that point.  I pulled the master and it worked.  I guess it might be related to https://github.com/mailpile/Mailpile/issues/1133

JackDca:
Hmmm. Were you using GnuPG 1.4 or 2.1? I've always used 2.1 with no problem, and I've seen 1.4 do some strange things.

JackDca:
The updated PR should now work.

bjmgeek:
I was using 2.1

bjmgeek:
That worked.  I was able to select my current key.

JackDca:
Thanks for taking the time to test it!

BjarniRunar:
Thanks both of you for your work on this. I've merged @JackDca 's fix and am closing this issue - please feel free to reopen or file a new one if the problem resurfaces.

PR #2145, opened 2018-09-24
Fix setup dialog offering revoked key (#2142)

JackDca:
This PR adds screening of the existing keys that are considered to be listed in the Add Account-Create a new Account-Security and Privacy-Encryption pull down list, so that disabled, revoked and expired keys are excluded, fixing #2142.

- In mailpile.crypto.gpgi.GnuPGRecordParser._parse_keydata() a boolean attribute "expired" is added and set to True IFF the "e" flag is present in the "validity" field.
- In mailpile.crypto.gpgi.GnuPG.list_secret_keys() the "expired" attribute is added to the attributes that are copied to the parsed output of the gpg --list-secret-keys command, from the parsed output of the gog --list-keys command. This coying is a workaround for a problem with gpg version < 2.1 in which the --list-secret-keys command does not display all the needed validity information. It is unnecessary for gpg version >= 2.1.18.
- In mailpile.plugins.crypto_gnupg.GPGKeyListSecret, an option --usable is added to the crypto/gpg/keylist/secret command, which causes disabled, revoked and expired keys to be omitted from the command results.
- In shared-data/default-theme/html/profiles/account-form.html, the --usable option is added to the call to crypto/gpg/keylist/secret.
 
Note: The crypto/gpg/keylist/secret command SYNOPSIS formerly listed an argument "address" but that argument was not actually implemented. This may be because it is easier to implement the screening of the key User IDs for the correct email address in account-form.html than to pass the email address from the html interface back to the crypto/gpg/keylist/secret command.

This has been tested on the current Mailpile master, on Debian stretch, using gpg 2.1.18, with a set of test keys including disabled, revoked and expired keys.


BjarniRunar:
This looks great! Thank you!

PR #2146, opened 2018-09-26
Allow missing MDC info if MDC is not required (issue 2075)

vezult:
Looks like this has been stalled since July. Does this patch satisfy your concerns?


BjarniRunar:
Thank you for doing this! I will take a look as soon as I have some time.

BjarniRunar:
I have merged this in my local tree, will push up with my next batch. Thank you for doing this!

Issue #1991, opened 2018-01-27
Copy/fork the spambayes Classifier class (and dependencies)


BjarniRunar:
In order to simplify future porting to Python 3.x (Issue #160) and to avoid an external dependency on an unmaintained library, we should probably just fork the `spambayes.Classifier` class and copy it (and its dependencies) into our tree.

I propose we call it `mailpile.spambayes.Classifier` - to give credit where credit is due, if nothing else.

The port/fork should make a point of deleting as much code as possible, but leave copyrights and credits intact. There are all sorts of things in there, including URL retrieval (!) and a tokenizer, which we are not using. Less code means less maintenance, fewer places for bugs to hide and less work porting when the time comes.

deepaknalore:
Can I take this up? 
@BjarniRunar Can you please give me more insights on how to go about this?

BjarniRunar:
I'm very sorry about the late response; If you @deepaknalore, or anyone else are still interested; I don't have any clearer ideas about how exactly to do this, than those stated in the initial issue.

Basically, Mailpile is only using a tiny fraction of spambayes. This issue basically requires just copying as little as possible from the spambayes source into Mailpile's own tree and then porting it so it a) works and b) is compatible with Python 3.x.

Steps a) and b) don't have to happen at the same time. And probably shouldn't.

vezult:
Just to see what this would entail, I stripped everything out of spambayes except for the classifier module and its direct dependencies:

https://github.com/vezult/spambayes/tree/strip_for_mailpile

I also tried to make it work with python3, avoiding any invasive changes. I'm sure this means that my approach in some cases leave something to be desired, but it seems to have been effective. The tests seem to run successfully in Python 3.5.3, and Python 2.7.13, in my testing (`python Tester.py`).

If I attempt to import this work into mailpile, would you rather I import the original files, and then apply my patches to get them working, or import the files in their current "working" state, and do any additional changes from there?

BjarniRunar:
Hey @vezult that's awesome! Thank you!

I think my preference would be to import them in their original state and then commit your changes, just in case someone wants to go exploring, and then add integration commits on top of that.

Note that committing this to the Mailpile tree directly is the lazy short-cut; the more community friendly way would be to create a new package in PyPI so other projects could make use of it too. Bayesian classifiers are widely applicable and useful things! But I am 100% OK with taking the short-cut, if someone wants to factor it out into a project of its own later that's not exactly hard.

[...]

PR #2150, opened 2018-10-02
Issue #1991  import of spambayes classifier

vezult:
Added spambayes classifier to the mailpile tree.

  * Imported only the classifier and its dependancies
  * Removed external URL retrieval code
  * Modified code to be python2.7/3.x compatible
  * Generally tried to modify everything as little as possible


vezult:
Testing:
    * All spambayes tests pass
    * Autotagging remains enabled even in a virtualenv without an external spambayes module
    * No debug/error messages from the autotag plugin
    * New messages show up in the spam folder

BjarniRunar:
Thank you for working on this! I'm going to finish my current round of Autocrypt and PGP-key related work before I dive into this, so there will be a bit of a delay before this gets merged.

mailpile/spambayes/OptionsClass.py
```diff
@@ -89,6 +89,8 @@
     except ImportError:
         import io as StringIO
 
+    xrange = range
```
jonathan-s: Not sure how many similar cases like this there are. But an idea is to have a `py3.py` (or other naming) file and from that file you import all the the methods that have changed. 

That way it'll all be collected in one single file rather than a bit spread out.

jonathan-s:
Personally I would also squash all the python compability changes into one. Ie having "remove url retrieval" and "import spambayes classifier" as separate commits is fine, but the granular commits for each of the compability changes I would squash.

mailpile/spambayes/OptionsClass.py
vezult: The proper thing to do would probably be to use [six](https://pythonhosted.org/six/) which is listed in requirements-with-deps.txt. Seems like there is a desire to avoid external dependencies as much as possible, so I don't know if it would be acceptable to turn that into a direct dependency or not. I'd be for it.

It appears it's only use in this case would produce a 128 integer array, and it would only happen on import. Using a generator there doesn't seem like a huge win. I could just convert it to use `range()`. My goal in all cases was to try to stay as close to the original as possible. I don't know how much that matters.

vezult:
OK, I've:
  * squashed python3 compatibility commits
  * switched to use of [six](https://pythonhosted.org/six/) for handing between python2/3 (@BjarniRunar opinions?)
  * removed the spambayes dependency on [python-lockfile](https://pypi.org/project/lockfile/) in favor of [fasteners](https://pypi.org/project/fasteners/), which appears to already be a direct dependency

mailpile/spambayes/OptionsClass.py
BjarniRunar: There is a desire to avoid external dependencies that complicate packaging work. I don't think six is a problem, it should be everywhere by now.

BjarniRunar:
I have merged this in my local tree, once I've verified it seems to behave, I'll push to master. Thanks again for doing this!

BjarniRunar:
This is merged and public. I guess my rebasing confused GitHub, so it didn't recognize the merge. But it's done. Thank you!

PR #2155, opened 2018-10-14
WIP: store OpenPGP session keys in metadata index and use when available

### LINKED ISSUE #2154: Add an option to store PGP session keys
Use case: given a smartcard (#926) without backup, be able to read encrypted mail after smartcard has been destroyed or lost. Additionally, this would allow to delete local keys (e.g. to prevent from being used for signing as an additional measure to revocation certs, or to prevent from reading future messages encrypted for that key) without losing history.

As of current master, Mailpile will persist decrypted message until restart, and also update search index with decrypted message.

If this feature can be accepted under a flag, I would like to try to implement it. Is [`mailpile.search.read_message`](https://github.com/mailpile/Mailpile/blob/babc3e5c3e7dfa3326998d1628ffad5b0bbd27f5/mailpile/search.py#L1227) the relevant function?

yanalunaterra:
Resolves #2154.

This is a long way before being usable, and also requires an index migration.

### DIFF
--- a/mailpile/crypto/gpgi.py
+++ b/mailpile/crypto/gpgi.py
@@ -199,6 +199,9 @@ class GnuPGResultParser:
                 while keyid in encryption_info["have_keys"]:
                     encryption_info["have_keys"].remove(keyid)
 
+            elif keyword == "SESSION_KEY":
+                encryption_info["session_key"] = data[1].strip()
+
             elif keyword == "VALIDSIG":
                 # FIXME: Determine trust level, between new, unverified,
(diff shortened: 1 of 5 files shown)


BjarniRunar:
Neat!

Note that there is a "free-form" extensible section in the metadata; the body info (snippet filed) is (optionally) a JSON dictionary that can contain arbitrary data. We're already storing information about attachments there, using that to store the session key would also be appropriate. If done this way, you don't need to change the structure of the metadata index.

PR #2175, opened 2018-11-14
Offer all usable keys with desired email address when creating account

JackDca:
It was mentioned in #2142 that, when creating a new account, Mailpile offered just one existing PGP key to the user, even when more than one private key on the keychain had a user ID containing the desired email address. This PR fixes that behaviour.

- The existing keys in the `Create a new Account/Security and Privacy/Encryption key` pull-down are sorted in reverse order of creation date so that the newest is first.
- All keys with user IDs containing the email address selected in `Create a new Account/Basic Details` are made visible.
- The first (and therefore newest) key is the one that is initially selected in the pull-down.

Since the same code is used to generate the `Edit Profile` window, the order of keys presented there is also as described. However, in that window, all keys with matching user IDs were initially displayed and the already configured key was the one that was highlighted; that has not changed. The new code does come into effect if the email address is changed in `Edit Profile/Basic Details`.


BjarniRunar:
Excellent! This looks great, thank you.

Issue #2176, opened 2018-11-17
Mailpile docker image doesn't build on buster


mgax:
The Dockerfile in the repo tries to build an image based on debian "buster". It fails because of a missing package - `spambayes`:

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 mailpile : Depends: spambayes but it is not installable
            Recommends: tor but it is not going to be installed
            Recommends: pagekite but it is not going to be installed
            Recommends: python-stem but it is not going to be installed
            Recommends: wamerican-small but it is not going to be installed
            Recommends: gnupg-curl but it is not installable
E: Unable to correct problems, you have held broken packages.
```

Replacing `FROM debian:buster-slim` with `FROM debian:stretch-slim` fixes the problem and the image builds successfully.

BjarniRunar:
Thanks for the comments. Replies:

1. See my comment on the PR regarding Tor.
2. Recent versions of Mailpile will (should?) by default set the GnuPG home to the Mailpile data folder, which should resolve this, unless the user manually changes the settings on the Privacy page. Are you sure this is still an issue?
3. I think the user should be allowed to use the interactive setup themselves. Automating things may be appropriate in certain environments, but should not be the default in our Dockerfile.

mgax:
>     2\. Recent versions of Mailpile will (should?) by default set the GnuPG home to the Mailpile data folder, which should resolve this, unless the user manually changes the settings on the Privacy page. Are you sure this is still an issue?

Some files get created in that directory, but it looks like gnupg's internals, and it looks like it's safe to delete them. I can just remove the docker volume for `.gnupg` then.

PR #2177, opened 2018-11-18
Fix dockerfile

mgax:
Fixes https://github.com/mailpile/Mailpile/issues/2176 by changing the base image to stretch; it also fixes the entry point.


mgax:
> Can you revert the Tor service change?

The original code doesn't work, I just rewrote it keeping the original intent, I think.

> Current versions of Mailpile know how to launch and run Tor themselves

Then the dockerfile should not attempt to start tor at all, correct?

BjarniRunar:
Thank you for doing this! I agree with splitting things up.

Issue #2168, opened 2018-11-04
Improve Mailpile's notifications and event log


BjarniRunar:
The Mailpile notifications are too noisy/spammy, and there is some low-hanging fruit that could also make them more useful. A few things which should be done to improve this:

   * [ ] Suppress mail-source notifications more aggressively
   * [x] Find and fix the cause of superfluous time-outs at the API level
   * [ ] Improve the event log page to allow undo from there
   * [ ] Improve the event log page to give more useful details (instead of raw JSON blobs)

The notifications communicate extremely important information much of the time, but the spam is reducing the visibility of these messages and annoying users (to the point that people demand the feature be disabled entirely, #2166, which remains a hard no until someone proposes an alternate UI that accomplishes the same things).

The mail source notifications in particular are tricky; the information presented is extremely useful right after account setup, but quickly becomes distracting spam after that. The current (weak) workaround is these notifications reappear if you visit the account overview page, but if dismissed on any other page (search results etc.) they stay dismissed. They should at the very least auto-dismiss themselves after a while, but it's likely that an even better solution could be found (possibly not involving notification popups at all).

Marking this as Low Hanging Fruit, because it should be easy to make progress on individual components of this, even if the entire task is quite large.

BjarniRunar:
Hello @derekyyyan !

Thanks for choosing Mailpile for your project! I look forward to your contributions and will do my best to give helpful feedback.

Note that the tasks on the list above vary a great deal in size/complexity. The first two will probably be nearly one-line changes, but may require a fair bit of investigation to figure out what needs to be done. The latter two are more straightforward "just add new code to do new stuff" tasks.

In general, please note that it's much easier for me to review and merge if pull requests are small in scope - the first three issues would probably justify one pull request each, but the fourth point (implementing the notifications UI) would probably best be broken up into multiple changes, one for each type of notification that is being improved.

Good luck with your work, and thanks again! If you have any questions, our IRC channel (#mailpile on Freenode) is a good place to chat and get quick answers (assuming I'm awake and online). And the contributing guidelines (https://github.com/mailpile/Mailpile/blob/master/CONTRIBUTING.md) and the DEV FAQ should also be helpful.

BjarniRunar:
Addendum: In my last comment I probably overstated how small the a code change would be needed to address the first item on the list. Sorry! Please don't take my estimates too literally, they're just to give you an idea of what to expect - and one should also always expect surprises and wrong estimates. :wink:

[...]

PR #2187, opened 2018-12-01
Fix Excessive Timeouts

lee7670:
Fixes excessive timeouts by adding a 5 second buffer to the ajax request timeout and basing the timeout on the given wait time to account for server overhead when no new events are found.

Based on #2168 Part 2.
The problem was that the ajax timeout set for the GET request for the regular polling of event logs was exactly the same as the wait time specified in the request, but due to server overhead, actually took a little bit more time than the wait time when no events were found. This caused an AJAX timeout whenever no events were found. 
This was fixed by making the timeout not hard coded and instead scales based on the given wait time and now has a 5 second buffer to account for server overhead. This was tested by having Mailpile open for an hour and looking at the number of timeouts shown in chromes developer console. Over the course of an hour there were no timeouts compared to the previous implementation which timed out frequently.

### DIFF
--- a/shared-data/default-theme/html/jsapi/global/eventlog.js
+++ b/shared-data/default-theme/html/jsapi/global/eventlog.js
@@ -61,11 +61,15 @@ EventLog.poll = function() {
   // 3) Other tabs may rely on us putting things in localStorage that we
   //    ourselves don't care about.
   //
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Excellent work! Thank you!

Issue #2081, opened 2018-05-19
Reworking tests so they pass on Windows (and *nix)


noahhaasis:
The tests of the following functions will not pass because they assume unix paths:
DirCheck     
FileCheck    
NewPathCheck 
PathCheck
These are all in mailpile/config/validators.py

noahhaasis:
I thought a bit about possible solutions and the easiest that I can think of would look something like this:
from sys import platform 
....
"""
>>> on_windows = platform == 'win32'
>>> path = 'C:\\Users\\..\\' if on_windows else '/etc/../'
>>> abspath = 'C:\\' if on_windows else '/'
>>> PathCheck(path) == abspath
True
"""
I couldn’t test it yet because I currently don’t have access to a computer. I just wanted to throw this in to see what you think of this. Does it complicate the tests to much?

BjarniRunar:
That's actually relatively readable - I think that's a good approach. I'd merge that PR! :smile:

noahhaasis:
Great. I'll start working on this

noahhaasis:
TIL: Escaping a backslash in a doctest results in a error. They don't need to and mustn't be escaped.

oedbro:
I'd love to take a crack at this

PR #2191, opened 2018-12-08
Reworking Tests so that they pass on Windows and Unix

prateek1192:
#2081

### DIFF
--- a/mailpile/config/validators.py
+++ b/mailpile/config/validators.py
@@ -1,6 +1,7 @@
 import os
-import socket
+import platform
 import re
+import socket
 
 try:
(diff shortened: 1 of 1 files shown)


BjarniRunar:

BjarniRunar:
This patch doesn't make sense to me, and appears to be at least twice as long as it needs to be.

AFAICT, all it accomplishes is to not run os.path.expanduser() on MacOS, which seems like a regression to me.

Thanks for working on this, but since I don't understand what you're attempting (you did not explain) and the patch itself seems wrong, I'm going to close. Feel free to speak up if you think I'm wrong!

Issue #2129, opened 2018-07-30
Cannot see newly-imported key (even after clicking "Find Encryption Keys")


ryantate:
Needed to reply to someone added in to a thread. This person has a public key matching the email address on the thread.

First I tried "Find Encryption Keys" thinking it might check keyserver. It says it is looking but never returns.

I give up and find the key myself on the command line and import it into my main keyring managed by gpg. (Find it very quickly on the keyserver.)

I go back to mailpile and try to reply again. Mailpile does not realize I have this person's key now. Hmm. Seems like a bug. Anyway, I click "Find Encryption Keys" again and again it just searches and searches seeming to go into eternity with no results. Hmmm.

Is this expected behavior? When can I expect mailpile to discover keys after they have been added to the gpg keyring? Is there a time delay? Also, what does "Find Encryption Keys" do? I tried looking in the FAQ and googling around for all this, all I found was this old Issue with a similar problem, which was closed because it was believed fixed. https://github.com/mailpile/Mailpile/issues/1048

Thanks for any help. (I am on 1.0rc3)

JackDca:
This looks like another instance of #2111.

ryantate:
Unless I'm missing something, which I may be, #2111 is specific to importing keys from mail attachments. 

In this case, the mail was simply added directly to the gpg keyring via the gpg command line tool. (A keyserver was consulted via gpg --search-keys, but this is incidental.)

My understanding is that this is supposed to just work -- mailpile pulls keys from the gpg keyring.

Is there something in your code that would cause gpg keyring not to be consulted fresh every time a new mail is made? Is this information cached somewhere? Am I expected to restart Mailpile every time I add a new key, or no? Etc. Since the expected behavior is not documented in the FAQ I am not sure if this is a bug or what.

JackDca:
In my opinion this is definitely a bug. You are correct about the difference between this case and #2111, but I think it's the same bug.

I am NOT  an authority, but I have been trying to puzzle out this behaviour, here's how it looks to me:

It appears that the key fingerprint must be in the addressee's record in Mailpile's vCard store in order for Mailpile to recognize that it is able to send mail to that addressee (as signified by the green lock icon to the right of the address in the "To" field of the compose window).

[...]

PR #2192, opened 2018-12-12
Fix web interface key import from keychain - part of #2129

JackDca:
In #2129 some problems were reported with importing a key that has been added to the keychain using gpg from the command line, so that it is on the keychain but is not in Mailpile's VCards. A key must be identified on a VCard for a contact in order for Mailpile to use it for encryption. PR #2182 fixed the underlying problem with the `crypto/keyimport` CLI command, but there was still a problem when such a key import was attempted from the web interface. Specifically, a search for keys using the addressee's email address found the key, but the `Import Key` button was not displayed so the user had no way to import it.

The problem with the web interface was that `template-crypto-encryption-key` in `shared-data/default-theme/html/partials/hidden.html` checks the result of the `crypto/keylookup` command for the attribute `on_keychain`. If it is `True`, the `Import Key` button is not displayed.

This PR fixes that problem by adding an attribute `in_vcards`. If either `on_keychain` or in`_vcards` is `False`, the Import Key button is displayed.

This is a little more complicated than it at first appears (of course!!) because it is then necessary to set the `in_vcards` attribute to the correct value when a key is first found by `crypto/keylookup` and then to update it (and `on_keychain`) when the key is imported by `crypto/keyimport`.

The message after a successful import, `This key is on your key chain` was changed to `This key is available for use` to more accurately reflect what has happened. (In any case a casual user may notknow what a keychain is!)

There is a conceptual issue in that `in_vcards` is not a property of the key alone, it depends on the email address for which a key is required. The key could conceivably be in some VCard other than the one that is relevant. Nevertheless, this fix appears to work satisfactorily.

### DIFF
--- a/mailpile/plugins/keylookup/__init__.py
+++ b/mailpile/plugins/keylookup/__init__.py
@@ -53,7 +53,6 @@ def _update_scores(session, key_id, key_info, known_keys_list):
             score = 9
             reason = _('Encryption key has been imported')
 
-        key_info["on_keychain"] = True
         key_info['scores']['Known encryption keys'] = [score, reason]
 
     # FIXME: For this to work better, we need a list of signing subkeys.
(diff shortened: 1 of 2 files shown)


BjarniRunar:

BjarniRunar:
Thanks for doing this!

Issue #2159, opened 2018-10-25
Implement native PGP keyserver lookups (w/o GnuPG)


BjarniRunar:
Currently Mailpile relies on GnuPG to perform keyserver lookups. The keyserver lookup protocol is HTTP(S) based and relatively simple; we should handle this ourselves so the lookups conform to the same policies as the user has requested in their privacy settings (regarding Tor etc), and so the user can see the network access in the Mailpile network log.

The code currently lives in `mailpile/plugins/keylookup/__init__.py` and there are pretty clear FIXMEs explaining what is needed, and plenty of examples elsewhere in the code (e.g. in the `wkd` module) of how to perform HTTPS lookups from within Mailpile. Note that the HKP response protocol appears to follow the same syntax as GnuPG's `--with-colons` output, and there's a parser for that in the `mailpile.gpgi` module. So don't go writing your own parser! :wink: 

This would make a good task for a new contributor.

omriluz:
is anyone tackling this? I would like to try!

BjarniRunar:
@omriluz Nobody else has stepped up AFAIK, so go right ahead! If you have any questions, I should be relatively available tomorrow afternoon (UTC) on #mailpile on Freenode IRC.

BjarniRunar:
Note that I have updated the main issue text with a few more hints on how to get this done.

philratcliffe:
As there hasn't been much activity in this issue for a while, I've added  PR #2193 for consideration. Apologies if it is being worked on @omriluz and happy for my PR to be closed if this is the case.

dvtkrlbs:
If there is no activity on this issue i can try to solve it. 
Disclaimer: This will be my first kind of contribution so i might ask a lot of things 😁

BjarniRunar:
This is done, thanks @philratcliffe!

PR #2193, opened 2018-12-14
Add native PGP keyserver lookups (Issue 2159).

philratcliffe:
Updates so that Mailpile performs PGP key server lookups directly rather than through GnuPG.


BjarniRunar:

philratcliffe:
No worries @BjarniRunar; thanks for creating such a great email client!

fiza1160:
#2159

mailpile/plugins/keylookup/__init__.py
```diff
@@ -513,51 +518,92 @@ class KeyserverLookupHandler(LookupHandler):
...
+    PRIVACY_FRIENDLY = True
     PRIORITY = 200
     SCORE = 1
 
+    # During testing, there were frequent HTTP gateway errors returned from
+    # hkps.pool.sks-keyservers.net so sks-keyservers.net was added too.
+    KEY_SERVER_BASE_URLS = [
+        "https://sks-keyservers.net/pks/lookup",
+        "https://hkps.pool.sks-keyservers.net/pks/lookup",
+    ]
```
wiktor-k: In my experience https://keyserver.ubuntu.com/ has the highest uptime and lowest response times.

BjarniRunar:
Just a note: Merging this has become higher priority for me, since it will allow Mailpile to put upper limits on key size and thus protect GnuPG from SKS keyserver spam, as described in https://gist.github.com/rjhansen/67ab921ffb4084c865b3618d6955275f

For now we've reconfigured Mailpile in master/nightly to avoid the SKS keyservers entirely, but I think it would still be valuable to our users to be able to fall back to the keyserver pool as a last resort, simply because all the modern replacements are still at an early stage of development.

I'm going to try and find time to look at this, this week. Sorry again about the long delay.

Issue #2204, opened 2019-01-25
Treat certain URLs as attachments in the UI


BjarniRunar:
It is becoming increasingly common for people to send around attachments as URLs instead of attaching files. It would be nice if Mailpile recognized such things and presented in a friendly way in the UI. The groundwork for this feature was laid in ac03b658c1a50672102dccd205e49739f709177e, but it is incomplete.

Progress and proposed tasks:

   * [x] Change the `attachments` search result view to include "URL attachments"
   * [ ] In `mailpile.search`, scan text parts for URLs (there is a FIXME)
   * [x] In `mailpile.search`, scan HTML parts for URLs
   * [ ] Compile a list of trusted, known file sharing site URL regexps (Google Drive, etc)
   * [ ] Set sane `prefs.attachment_urls` defaults in `mailpile.plugins.setup_magic`
   * [ ] Consider whether the UI of message display also needs changing
   * [ ] Add a UI behaviour allowing a user to click on a link and say "treat URLs like this as attachments" to add to the `prefs.attachment_urls` list.

BjarniRunar:
Anyone who wants to add examples of well-known, trusted file sharing site URLs in the comments here is encourage to do so. Google Drive? DropBox? Do things like OwnCloud have a pattern we can recognize?

BjarniRunar:
Note: Before anyone requests Mailpile go and download such things in the background, rather than delegate to the user's browser; I don't think that is possible. These URLs usually lead to a human-friendly web interface of some sort, they are rarely a direct download link. The UI may also require the user authenticate or log in. The best we can do is recognize and improve the presentation of these URLs (and make them searchable).

byaka:
think Github also can be trusted URL

RS357:
Hey, I'd like to work on the second tick box task - scanning text parts for URLs. Should we have a chat on IRC about how to go ahead with this? Cheers!

byaka:
> Hey, I'd like to work on the second tick box task - scanning text parts for URLs. Should we have a chat on IRC about how to go ahead with this? Cheers!

Think u can make one in discord or something and just link here :)

BjarniRunar:
@RS357 Sure! Just join #mailpile and if I'm around we'll chat. But I must warn you, I'm often busy elsewhere so it may take a little while for us to sync up. If you leave your client connected you'll see my replies even if we don't manage to be online at the same time.

qxtaiba:
[...]

PR #2215, opened 2019-04-08
Textpart scanning for valid url and list of trusted urls of common file sharing sites

deepak21188:
We are a group of students. We studied the architecture of Mailpile as a part of software architecture and documentation course. We are doing this pull request as a part of our assignment. The pull request is regarding the issue https://github.com/mailpile/Mailpile/issues/2204. We have tried to complete certain tasks mentioned in https://github.com/mailpile/Mailpile/issues/2204#issue-403099577. List of tasks done as follows:
1. Resolved FIXME: Scan text parts for URLs, in search.py
2. Created a list of trusted URL regexps of common file sharing sites
3. Set prefs.attachment_urls defaults in mailpile.plugins.setup_magic


BjarniRunar:
Thank you for doing this!

I have to warn you it may take me some time to get a review done; I have been taking some time off from this project and there is a bit of a back-log of issues and PRs that need my attention. My first impression is that you've done good work here, but please be sure your supervisor understands that I may not be able to respond right away.

It would be a shame if you missed out on credits because I'm unresponsive.

deepak21188:
Thank you so much!  The guideline for the assignment says that the pull request does not have to be accepted for grading, hence, we are not worried about that.

PR #2216, opened 2019-04-21
Searches textpart for URLs and appends to urls

nkfarwell:
This is my first pull request. Any feedback is greatly appreciated. Addresses the `FIXME` and deletes that comment.

### DIFF
--- a/mailpile/search.py
+++ b/mailpile/search.py
@@ -1258,6 +1258,11 @@ class MailIndex(BaseIndex):
                     ctype = 'text/html'
                 else:
                     # FIXME: Search for URLs in the text part, add to urls list.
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Sorry about the long delay in responding. Thanks for attempting this!

However, just from looking at the code, I know it cannot possibly work. The `re.findall` function returns a list, which you're appending directly (as a list, not as the items in the list) to the list of URLs. Using `urls.extend` might work... but this is just obviously completely untested code. That's really not OK.

You *must* test your code before you submit it for review, anything less is disrespectful towards the people who spend time responding and reviewing. I'm sure you can do better! :grin:

PR #2224, opened 2019-07-10
Run futurize --stage1

yegortimoshenko:
`futurize` produces code that is compatible between Python 2 and 3. `--stage1` leverages `__future__` without requiring any external deps. See: http://python-future.org/automatic_conversion.html

To reproduce the exact same tree, one can run:

```
pip3 install future
echo > mailpile/tests/test_mail_generator.py
futurize --stage1 -nw .
git checkout -- mailpile/tests/test_mail_generator.py
```

This skips `mailpile/tests/test_mail_generator.py` for two reasons: first, it contains a non-UTF8 compliant symbol (by design), second, this file is already Python 3 compatible.

Want to keep these changes as reviewable as possible. I'll send a few more patches (`rfc822` -> `email`, syntax fixes for Python 3 that `futurize` didn't pick up) next.

### DIFF
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import getopt
 import gettext
 import locale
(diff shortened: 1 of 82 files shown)


yegortimoshenko:
For the record, I've tested this changeset locally against my own Mailpile installation, specifically: sending/receiving messages, encryption, search.

BjarniRunar:
This relates to #160. Thanks for doing this. :heart:

BjarniRunar:
This is mostly merged in my local tree, will test a bit and explore before pushing to master. One thing I had to fix (so don't go redoing it) is the use of print() in doctests, looks like futurize overlooked those.

yegortimoshenko:
Thanks! Great.

I'll be sending `rfc822 -> email` patchset next, it's so small it shouldn't conflict with anything.

BjarniRunar:
This is merged, but I guess my rebasing changed enough things that Github didn't recognize it as such. Thanks for doing this!

PR #2225, opened 2019-07-10
Make keybindings maintain focus

yegortimoshenko:
This patch significantly improves usability of keybindings by ensuring that focus between changes is not reset.

Preview

### DIFF
--- a/shared-data/default-theme/html/jsapi/ui/keybindings.js
+++ b/shared-data/default-theme/html/jsapi/ui/keybindings.js
@@ -26,24 +26,11 @@ Mailpile.keybinding_move_messages = function(op, keep_new) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


BjarniRunar:
I have not tested this, so forgive me if I'm getting this wrong.

Note that the inconsistency between different actions is deliberate. Some tags are "containers" and we move messages (thus removing them from the view), others like read/unread are attributes which just change the display of the message unless you are explicitly searching on for messages that are read/unread (in which case they do disappear from the view).

We can debate whether it's a good idea (maybe it is not), but if I'm reading this right, your change means the keybindings now behave differently from the buttons in the selection-action area, which I consider a regression.

BjarniRunar:
Maybe if you could better explain what you're trying to fix (the description does not match the code change IMO, so I'm missing part of your intent), I could give better feedback. :)

PR #2226, opened 2019-07-11
Provide error code with attachment upload err

### LINKED ISSUE #2220: Improve HTTP Error code messages in UI
When attempting to uploade images that are too large the backend correctly returns a 413 Entity Too Large error, but the frontend displays a very generic error message. 

Returning the HTTP message would be pretty helpful :)

Screenshot 2019-06-04 at 19 00 25
Screenshot 2019-06-04 at 18 59 46

Pectojin:
err.message is very generic on attachment upload failure `HTTP Error.`.

Adding `err.code` should give at least something duck-able so the user can troubleshoot more easily.

Fixes #2220

### DIFF
--- a/shared-data/default-theme/html/jsapi/compose/attachments.js
+++ b/shared-data/default-theme/html/jsapi/compose/attachments.js
@@ -200,7 +200,7 @@ Mailpile.Composer.Attachments.Uploader.init = function(settings) {
         }
       },
       Error: function(up, err) {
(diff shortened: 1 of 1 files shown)


BjarniRunar:
Thanks for doing this! I think that error message could probably be improved (it's quite bad from an i18n point of view, for one thing), but it was already broken and your change makes it less so. :smiley:

Issue #2211, opened 2019-03-01
Make Tor a soft dependency, cope gracefully if tor hangs


BjarniRunar:
Apparently, the `tor --version` command can silently hang in some configurations. This prevents Mailpile from starting up, as discussed here: https://community.mailpile.is/t/mailpile-just-hangs/53

We should fix that!

PR #2232, opened 2019-08-07
don't exit when tor binary check fails

jo285317:
[Make Tor a soft dependency, cope gracefully if tor hangs #2211](https://github.com/mailpile/Mailpile/issues/2211)

This doesn't break the app, as platforms.BINARIES.Tor will be still be false, and thus there will be no way to attempt to load tor.

### DIFF
--- a/mailpile/app.py
+++ b/mailpile/app.py
@@ -4,6 +4,7 @@ import gettext
 import locale
 import os
 import sys
+import warnings
 import traceback
 
 import mailpile.util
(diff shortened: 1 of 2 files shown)


BjarniRunar:
Thank you for doing this!

I'll try and review the functionality soon, and get it merged if I don't find any show-stoppers.

In the meantime (and this is optional), it would be helpful if you could remove your use if the Python warnings. I realize we're probably lame for not using them, but they're not currently used anywhere else in the code and I would rather we avoid a proliferation of different logging strategies within the app. Switching to more standard Python logging would probably make sense, but it needs to happen in a coherent way, not like this.

If you don't find the time, I'll be stripping that stuff out myself when I merge.

Thanks again!

PR #2236, opened 2019-09-06
FIXES 2235: Changed timeout for executable search from 1 second to 5

### LINKED ISSUE #2235: Running mailpile on slow disks (ex. docker server) can't find executables
Ran into an issue getting Mailpile installed. I've downloaded the repo here using this dockerfile: [https://hub.docker.com/r/rroemhild/mailpile](https://hub.docker.com/r/rroemhild/mailpile) and build a custom container based on 1.0.0rc6.
Whenever I start a container with the image, mailpile fails to start up with errors similar to this:
```
Required binary missing or unusable: Tor not found

If you know where it is, or would like to skip this test and run Mailpile
anyway, you can set one of the following environment variables:

    MAILPILE_TOR="/path/to/binary"
or
    MAILPILE_IGNORE_BINARIES="Tor"

Note that skipping a binary check may cause the app to become unstable or
fail in unexpected ways. If it breaks you get to keep both pieces!
```

Even when the which command reports the binary exists. What is happening is that the timer on the binary check is expiring due to the time it takes the disks on the docker host to load the binary from the container on disk to memory and execute the program. I'm going to put up a PR to adjust the timeout to 5 seconds, just wanted to make the issue first to track it.

Specs on the system under test:
Ubuntu 18.04
Docker version 19.03.1, build 74b1e89
Mailpile 1.0.0rc6
7.5TB JBOD (asst. disks) with VMWare ESXi 6.5U2

Thanks,
Tuxprogrammer

Tuxprogrammer:
FIXES #2235: Changed timeout for executable search from 1 second to 5 seconds for slower computers.

### DIFF
--- a/mailpile/platforms.py
+++ b/mailpile/platforms.py
@@ -66,7 +66,7 @@ def DetectBinaries(
                     BINARIES[binary] = env_bin
                     continue
             try:
-                mailpile.util.RunTimed(1.0, _run_bintest, binary_test)
+                mailpile.util.RunTimed(5.0, _run_bintest, binary_test)
                 BINARIES[binary] = binary_test[0]
                 if (not os.path.dirname(BINARIES[binary])
                         and not sys.platform.startswith('win')):


BjarniRunar:
Thanks for doing this!

Issue #2080, opened 2018-05-19
Add Account should not permit duplicate email address


JackDca:
If the Add Account dialog is given the same name and email address as an existing account it will create a second account duplicating the email address of the existing account as requested. (In testing this the encryption key created during the setup of the first account was selected for the new account - the creation of a new key was for the new account was not tested.) Then, in the home screen on the line for the new account, if the Trash icon is clicked, the program appears to start removing the duplicate account, but the display stops updating.

Having the same email address attached to two different accounts in the same pile does not make sense and is sure to cause problems. The Add Account dialog should screen the supplied email address and reject it if it is the same as the email address for an existing account. Preferably the email addresses should be canonicalized before comparison.

BjarniRunar:
Agreed, thanks for filing this.

JackDca:
@deepaknalore - Although I reported this issue, a fix would require mods to the front end, which is not my strong area, so I don't plan to work on it. If you can provide a fix, I say go for it!

noahhaasis:
Since @deepaknalore hasn't commented anything yet I assume he doesn't currently work on this. I'd like to look a bit into the problem. Can somebody give me a hint where in the codebase I might have to look?

JackDca:
I don't speak HTML/JS/CSS but here is a  suggestion on how to get a grip on the code:

- The Add Account pop-up that presents the "E-mail" field to be filled in has a subtitle "Basic Details" which is a good target for a search because it is specific enough to avoid a lot of irrelevant hits.
- A global search of all the source for "Basic Details" finds `default-theme/html/profiles/account-form.html`.
- A search of that file for "E-mail" finds this definition:
```
        <div class="left" style="margin-right: 1em; width: 29em;">
          <label>{{_("E-mail")}}</label>
          <input type="text" name="email" style="width: 100%"
                 class='fpa-email fpa-basic-username' data-next="#basics-next"
                 placeholder="ada@example.com" value="{{ result.email }}"
                 onchange="javascript:_fpa.email_changed();">
        </div>
```
Hope this gets you started!

I use Eclipse to manage my source - it facilities various kinds of global and local searches. It integrates with git so you easily can compare against other versions. (There's also an Eclipse-based Python IDE which I have never learned to use!)

BjarniRunar:
Thanks @JackDca for the great comments, and thanks @noahhaasis for the offer to help! More hints:

A good solution to this would look something like this:

[...]

PR #2240, opened 2019-10-25
Release/1.0

SiddheshJoshi1012:
[AddUser.txt](https://github.com/mailpile/Mailpile/files/3771332/AddUser.txt)

Try adding this to your code to avoid Duplicate User name while signing up or adding a new user. This will raise a Exception when any duplicate user name is added. And prompt the user with a message "Duplicate User Id". For more security purposes you can return than user id to let the user know that this user id already exists.


BjarniRunar:
Hello, thanks for your efforts. However, I don't understand what you are doing here.

I don't know what problem you are solving, and the this pull request is so full of unrelated noise that I can't tell what is going on. Please work off the master branch, not the release/1.0.

SiddheshJoshi1012:
Actually you have put a issue for Duplicate user id. So i've tried to solve
that issue. As you said you don't understand it. I'll try something else.
Btw thanks for your views. :)

BjarniRunar:
Ah, I see! Usually it's good to reference the issue by number, like so: #2080 

Feel free to try again; the main thing is that if you're making changes, you really need to be working against the Github master branch. Otherwise the pull request gets cluttered up with unrelated things and becomes impossible to review, let alone merge.

That code snippet in your .txt file is obviously broken though, just looking at it I can see it won't run (you have three variables in your format string, only two in the tuple). Also, writing usernames and passwords to a file in clear-text is something we will never, ever do. So I think whatever you are doing still needs work.

Hopefully these comments are helpful.

Issue #1888, opened 2017-07-24
Password popup problems


BjarniRunar:
The password input dialog pops up and behaves in unhelpful ways.

   * [x] it pops up when the user has already entered a password but no new problems have occurred
   * [x] when it pops up, it will let the user forget their password but not enter a new one
   * [ ] when the user enters a password, it isn't immediately checked (which causes confusion)
   * [ ] the SMTP client (mail sending) notifications do not currently trigger password dialogs
   * [ ] if the SMTP credentials and IMAP credentials differ, the user may be SOL.

This all needs fixing.

JackDca:
I have encountered this bug. I cannot find any way to get around it. At present Mailpile appears to have forgotten the passwords to both my imap servers and I have note been able to find a way to log in. This is more than unhelpful, it is fatal!

The strange requests for password started after I did a pull from master recently. Not sure how long it had been since I did that, but the cause appears to be a change that was made fairly recently.

The IMAP settings dialogue used to include a field for the password. After the recent update, a separate dialog box was used to request the password. That appears to be when the problem started.

I am going to try to revert to an earlier commit.

JackDca:
Could this be related to  (cut from git log):

commit 46f5d32456d5453d660213d00ab67541d141bf2c
Author: Smári McCarthy <smari@occrp.org>
Date:   Wed Jul 19 16:19:14 2017 +0000

    Make login credentials pop up occasionally.

These pop ups are confusing and you can press a button which makes Mailpile forget the IMAP passwords. Looks like that is what happened to me.  I have recovered access to one IMAP server by copying mailpile.cfg.5 (a few hours old) to mailpile.cfg. The old .cfg file still contained the IMAP password for one of the account.

I don't think a casual user will want to be repeatedly asked to enter credentials.

BjarniRunar:
Is this still a problem? I would have expected it to not bother you for a while, but then prompt you for the passwords again later on, at the latest after a restart.

There is also a tool now for managing credentials (click Notifications or visit the Privacy page, and then select the Account Password Management tool from the toolbar (the icon is a person).

BjarniRunar:
(Above question is for @JackDca, I don't think the points I raised at the top of the issue have been addressed.)

JackDca:
[...]

PR #2242, opened 2019-11-06
Handle non-password related IMAP login errors appropriately re #1888

JackDca:
One of my ISPs sometimes give a response `System error` when an IMAP login is attempted. Usually a second login attempt after five minutes will succeed. (This is password authentication, not OAuth.) But, Mailpile treats this error as a bad IMAP password and will not attempt another IMAP login until the user enters a "correct" password. Confusing, since Mailpile already has the correct password. Assuming that the user has not memorized the password, it is then necessary to `Display Remembered Account Password`, copy it somehow, `Lock Account and Forget Password`, `Unlock Account and Remember Password` and enter the password. Arrrgh!!! This was described in [comments on #1888](https://github.com/mailpile/Mailpile/issues/1888#issuecomment-410461276).

This patch creates a new error type `login` with message `IMAP Login error`. The IMAP response is checked to see if the error is a password error. If not, the `login` error is used. Only if there is a password error is the existing `auth` type error with message `Invalid username or password` used, blocking further login attempts and triggering a request for a new login in the `Notifications` popup.

Unfortunately, detecting a password error is not straightforward. RFC5530 specifies standardized IMAP responses which are always upper case and enclosed by `[ ]`, but adherence to RFC5530 is not required. Of the three ISPs using password (not OAuth) authentication that I use, only one complies wtih RFC5530. In this patch, if the response contains any code in RFC5530 format, the code checks for RFC5530 codes only. If there is no RFC5530 format code in the response, it checks for `PASSWORD` (case independent). This works with the providers that I use.

Conceivably an ISP could respond to a bad user id or password with a non-RFC5530 response that did not contain `PASSWORD`. In that case, the user would see `IMAP Login error` in the notification window every five minutes, which should be enough to alert the user to the problem.

### DIFF
--- a/mailpile/mail_source/imap.py
+++ b/mailpile/mail_source/imap.py
@@ -685,8 +685,9 @@ def _connect_imap(session, settings, event,
 
         username = password = ""
         try:
-            error_type = 'auth'
-            error_msg = _('Invalid username or password')
+            error_type = 'login'
+            error_msg = _('IMAP Login error')
+            auth_error_type = auth_error_msg = None
             username = settings.get('username', '').encode('utf-8')
             password = IndirectPassword(
                 session.config,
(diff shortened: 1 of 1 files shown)


BjarniRunar:
This looks great, thanks for doing this.

JackDca:
:)

PR #2252, opened 2020-05-01
Fix for Issue #2045

wbenfie:
I made some changes on default.css and topbar.html to decrease the size of the icons on the top. Currently it is less buggy and all five items appear at all screen sizes.


BjarniRunar:
Thank you for doing this, and sorry about the late response - I have been taking a break from Mailpile work for the last few months. I'll try and merge this locally to take a look soon.

PR #2270, opened 2020-10-06
Actually install tor into container [WIP]

Leopere:
Tor is not installed but on lines 14 and 15 there is an attempt made to enable it on container startup which likely will not work correctly but at least now tor will be installed.  Containers are typically single process things but if Mailpile is running `tor --version` as mentioned in #2211 you will need it in the same container.  I don't believe anything calls init.d/systemd inside of containers so there may need to be an alternative path explored. 

The proper method would be to spin up a docker container all on its own and then either do a socket mount between the containers or do a "cross host[container]" connection from the Mailpile container to the Tor container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ENV GID 33411
 ENV UID 33411
 
 RUN apt-get update && \
-    apt-get install -y curl apt-transport-https gnupg && \
+    apt-get install -y curl apt-transport-https gnupg tor && \
     curl -s https://packages.mailpile.is/deb/key.asc | apt-key add - && \
     echo "deb https://packages.mailpile.is/deb release main" | tee /etc/apt/sources.list.d/000-mailpile.list && \
     apt-get update && \


sebcworks:
You may also use someting like supervisiord to manage several services inside 1 container

Leopere:
@sebcworks you can run several services inside of a single container but that's typically less than ideal.

sebcworks:
Of course, it is a bad practise and it's not what containers are designed for, but to do what is indented here, it is a "quick and easy" solution as it does not imply rewriting Mailpile code...

Leopere:
Fair enough. Yeah i almost think it might be better just to write a vagrant
file.

BjarniRunar:
Thank you for working on this! Please accept my apologies for not responding sooner.

However, I don't think it makes sense to install tor if it isn't going to be launched as well, so I won't be merging this as-is.

Leopere:
My suggestion was because if i recall correctly you don’t have tor in your container or a docker-compose file.

Sent with a Spark

PR #2278, opened 2020-12-07
Prevent duplicate emails or names in frontend

stefanb0305:
This PR will add a check for duplicate emails and names when filling out the Add Account form using the GUI. If the email or name is already used, then it will give the user an alert stating the problem - see the image below. The user will not be able to continue until they make the change.
We also added an alert if the name input box is left empty, which we believe should not happen.

image

### DIFF
--- a/shared-data/default-theme/html/profiles/account-form.html
+++ b/shared-data/default-theme/html/profiles/account-form.html
@@ -147,62 +147,77 @@
         },
         basics_next: function() {
           var email = $(pf + '.fpa-email').val();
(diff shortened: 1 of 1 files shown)


stefanb0305:
This is the first step to solving the issue mentioned in #2080

BjarniRunar:
Thank you so much for working on this, and please accept my apologies for not responding sooner.

This appears to overlap with #2279 - I will have to compare and see which I prefer, or combine the two if they are complementary.

PR #2279, opened 2020-12-07
Fix: add account should not permit duplicate email addresses

### LINKED ISSUE #2080: Add Account should not permit duplicate email address

mnahinkhan:
This PR fixes #2080.

The changes made were:

1. AddProfile in .../mailpile/plugins/contacts.py now checks the new email address before adding a new profile. (We make sure it is not an edit, as suggested in the Issue). So if creating a new profile, the API call fails if the e-mail address already exists.

2. The index.html template collects the accounts passed to it in results.profiles to extract and store the email addresses in a global JS variable.

3. The template .../shared-data/default-theme/html/profiles/account-form.html uses that array to perform an additional validation check on the email entered (we created a new error logo color for this class of errors).

So in summary, both the front end and back end have been updated to fix this issue. Please find screenshots illustrating the fix:

Screen Shot 2020-12-07 at 5 05 51 PM

Screen Shot 2020-12-07 at 5 06 56 PM

### DIFF
--- a/mailpile/plugins/contacts.py
+++ b/mailpile/plugins/contacts.py
@@ -1217,9 +1217,9 @@ class AddProfile(ProfileVCard(AddVCard)):
             all_emails = [ [vcl.value for vcl in vc.get_all('email')] for vc in all_vcards ]
             all_emails = [item for sublist in all_emails for item in sublist]
             
-            new_email = self.data.get('email', [None])[0]
+            all_emails = [email.lower() for email in all_emails]
+            new_email = self.data.get('email', [None])[0].lower()
 
-            
             if new_email in all_emails:
                 raise ValueError("duplicate email address used")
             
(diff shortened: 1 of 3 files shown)


BjarniRunar:
Thank you so much for doing this, and please accept my apologies for ignoring this for so long.

I will probably merge this to my local tree and push up along with other changes once I have had time to test it.

mnahinkhan:
Thank you! I would like to say that I had spotted a bug back then, which I thought to work on but didn't get the time to.

Unfortunately I have forgotten what the bug may have been. 

I know this is unhelpful but this is just to say I encourage you to test, as you stated you would 🙂.

In case you spot errors please feel free to state them here and I'll try to fix it (or if it's simple enough please feel free to fix of course). 

I'll also let you know in case it comes to my mind.

Issue #1476, opened 2015-12-23
Make the HTML viewer display in-line images


BjarniRunar:
The current HTML viewer can display remote images, but lacks the ability to display in-line (attached) images. Fixing this requires updating the Javascript (`mailpile/www/default/html/jsapi/message/html-sandbox.js`) to detect CID image URLs and convert them to attachment download URLs.

This should be relatively straightforward, marking as low hanging fruit. :-)

loidor:
Is this still an issue or am I doing something wrong? I can't get inline images to display at all.

BjarniRunar:
This is still an issue.

Since the HTML support has been completely insecure until very recently, fixing things like this wasn't a priority. This isn't on our roadmap for 1.0 (because priorities), but if someone does the work and sends a pull-request... :-)

loidor:
This is cleared by #1735.

Issue #2293, opened 2021-08-28
Message repetition when allowing image downloading


pisn:
Hello everyone,

I am new to the project, and while exploring and debugging I came across a previously undescribed bug: 

When the message has inline images to be downloaded and Mailpile asks if it is ok to display the messages. When you click "Okay, display the images" more than once, the body of the message gets repeated. If you keep clicking, it repeats indefinitely:

After one clicks:

![imagem1](https://user-images.githubusercontent.com/7537247/131231809-0ae0fc5a-7d1d-44a3-86fe-dc308ca220c6.png)

After two clicks:

![imagem 2](https://user-images.githubusercontent.com/7537247/131231812-e0f5c4b8-a002-4710-9d74-e5a96179e9f8.png)

After three:

![imagem3](https://user-images.githubusercontent.com/7537247/131231823-6a815cf9-6eeb-41f3-9614-33375ab80f87.png)

And so on...

I believe the best would be to make sure the image disappears. I will try to work on this on the next weeks.

PR #2296, opened 2021-09-06
Inline images 1476

pisn:
Hello everyone!

This patch fixes issue #1476 . 

The problem: inline images are not shown in Mailpile when they refer to an attachment ContentID rather than a web address. The "src" attribute of these images point to the attachment, but the browser cannot find this address. 

The solution: replace the "cid" in the "src" attribute by the base64 image source, contained in the same message downloaded from email server, but in a separated payload. This fixes the issue. 

The patch has been tested against emails containing inline attached images coming from gmail and outlook senders.

### DIFF
--- a/mailpile/mailutils/emails.py
+++ b/mailpile/mailutils/emails.py
@@ -932,6 +932,42 @@ class Email(object):
             result = self.msg_parsed
         if not result:
             raise IndexError(_('Message not found'))
+
+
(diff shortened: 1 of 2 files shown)


pisn:
I also added the correction for issue  #2293 

The problem was solved passing the "true" parameter of the SandboxHtml function when clicking one of the "accept" options in the display message.

PR #2305, opened 2021-12-04
Issue 1999, removed new email telemetry

JostonR:
issue 1999

Problem: Mailpile Welcome Home page does not refresh email statistics despite new emails coming in. I.e. the new email count and inbox count is wrong. 

Solution: Based on issue 1999, removing the telemetry numbers altogether seemed like an appropriate, initial response.


BjarniRunar:
Thank you for spending time on this, but I disagree with the change.

PR #2306, opened 2021-12-06
eased up suppression enforcements as per issue 2130

JostonR:
As per issue 2130, Mailpile is aggressively suppressing forwarded content chain mails even if the user hasn't yet seen the content. To counter this, we have taken away the hide classes on the content and signature, to continue to show the chain emails.


JostonR:
As per issue 2130, Mailpile is aggressively suppressing forwarded content chain mails even if the user hasn't yet seen the content. To counter this, we have taken away the hide classes on the content and signature, to continue to show the chain emails.

